PR #92, opened 2020-10-27
Make tests more resilient

MarkusH:

When the operator deploys a CrateDB cluster, the StatefulSet adds a
local PersistentVolume to each pod which can be used for debugging.
Historically, the operator used the "default" storage class for these
PVs. However, since "default" may map to block storages in the cloud,
e.g. on Azure or AWS, K8s would attempt to mount the same block device
onto multiple pods, which isn't possible. To mitigate the issue, the
operator's tests now ensure there's a local file system.

## Checklist

### DIFF
--- a/tests/test_operations.py
+++ b/tests/test_operations.py
@@ -48,9 +48,7 @@ async def is_cluster_healthy(conn_factory: Callable[[], Connection]):
 
 @pytest.mark.k8s
 @pytest.mark.asyncio
-async def test_restart_cluster(
-    faker, namespace, cleanup_handler, cratedb_crd, kopf_runner
-):
+async def test_restart_cluster(faker, namespace, cleanup_handler, kopf_runner):
     coapi = CustomObjectsApi()
     core = CoreV1Api()
     name = faker.domain_word()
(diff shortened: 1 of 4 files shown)


tests/test_operations.py
```diff
@@ -48,9 +48,7 @@ async def is_cluster_healthy(conn_factory: Callable[[], Connection]):
 
 @pytest.mark.k8s
 @pytest.mark.asyncio
-async def test_restart_cluster(
-    faker, namespace, cleanup_handler, cratedb_crd, kopf_runner
```
MarkusH, line 52: `cratedb_crd` is a dependency of the `kopf_runner` fixture.

tests/test_scale.py
```diff
@@ -73,7 +73,6 @@ async def test_scale_cluster(
     faker,
     namespace,
     cleanup_handler,
-    cratedb_crd,
```
MarkusH, line 76: `cratedb_crd` is a dependency of the `kopf_runner` fixture.

chaudum (approved):

PR #94, opened 2020-11-02
Support zone awareness on Azure

chaudum:

On the Node: The kubelet populates this with the zone information as defined by the cloudprovider. This will be set only if you are using a cloudprovider. However, you should consider setting this on the nodes if it makes sense in your topology.

-- [Docs: failure-domain.beta.kubernetes.io/zone](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#failure-domainbetakubernetesiozone)

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,10 @@ Changelog
 Unreleased
 ----------
 
+* Support Pod spreading across zones on Azure using weighted Pod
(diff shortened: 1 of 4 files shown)


MarkusH:
Looks good!

MarkusH (approved):

PR #103, opened 2020-11-06
Fix #84: Adjust parsing of replicas to use lower bounds

### LINKED ISSUE #84: Allow scaling down even when dynamic replicas have `-all`
In a dynamic replica setting, such as `1-all`, the lower bound counts for CrateDB to stay green. In this case, a cluster is considered healthy if there are still two nodes: 1 as the primary node and 1 with the one minimum required replica. The scaling operation should account for that.

MarkusH:
## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,12 @@ Changelog
 Unreleased
 ----------
 
+* Fixed parsing of replicas. Previously, in a replica settings like ``'2-5'``
+  or ``'2-all'``, the upper bound was used. This effectively made scale-down
+  operations impossible, at least for the ``'all'`` case. However, a table and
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -5,6 +5,12 @@ Changelog
 Unreleased
 ----------
 
+* Fixed parsing of replicas. Previously, in a replica settings like ``'2-5'``
+  or ``'2-all'``, the upper bound was use. This effectively made scale-down
```
chaudum, line 9:
```suggestion
  or ``'2-all'``, the upper bound was used. This effectively made scale-down
```

chaudum (approved):

Issue #75, opened 2020-09-18
Automatically push Docker image to public Docker registry on release


MarkusH:
It would be neat if a Docker image would be build and pushed to a public Docker registry (e.g. GitHub directly or DockerHub) upon releases.

PR #106, opened 2020-11-11
Add github action to build and publish docker images on new releases

Taliik:
This github action will trigger when a new tag is release and will build/publish/tag the docker image accordingly.
For a version like `1.0.0` or `1.0` there will also be a `latest` tag built, for versions like `1.0b1` there will not be a latest tag.

## Checklist

### DIFF
--- a/.github/workflows/docker_image.yaml
+++ b/.github/workflows/docker_image.yaml
@@ -1,0 +1,34 @@
+name: Build and publish Docker Image
+
+on:
+  push:
+    tags:
+    - "*"
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Docker meta
+        id: docker_meta
+        uses: crazy-max/ghaction-docker-meta@dd2e615
+        with:
+          images: crate/crate-operator
+          tag-match: '^[0-9]+\.[0-9]+(\.[0-9])?$'
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Build and publish
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64
+          push: true
+          tags: ${{ steps.docker_meta.outputs.tags }}


.github/workflows/docker_image.yaml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build and publish to registry
+        uses: elgohr/Publish-Docker-Github-Action@master
```
chaudum, line 16: I would nail the version to the current latests (2.22)

```suggestion
        uses: elgohr/Publish-Docker-Github-Action@2.22
```

https://github.com/marketplace/actions/publish-docker?version=2.22

.github/workflows/docker_image.yaml
```diff
@@ -0,0 +1,22 @@
...
+    tags:
+    - "*"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+
+    steps:
+      - uses: actions/checkout@v2
```
chaudum, line 13: Maybe also give the step a human readable name

.github/workflows/docker_image.yaml
```diff
@@ -0,0 +1,22 @@
+name: Build and publish docker image to docker hub
```
chaudum, line 1:
```suggestion
name: Build and publish Docker image to Docker Hub
```

chaudum (approved):

MarkusH (changes requested):
I'm not entirely sure from the changes which registry is going to be used. Is it Dockerhub? GitHub?

I think extending the installation instructions to include the corresponding `docker run` or snippet for a Kubernetes `Deployment` would be useful. At least to see what/how/where the image is and how it's referenced.

Please prefix the commit message with `Fix #75` to reference the commit and issue.

.github/workflows/docker_image.yaml
```diff
@@ -0,0 +1,39 @@
...
+    tags:
+    - "*"
+
+jobs:
+  multi:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
```
chaudum, line 14: I would prefer if we use the formatting we otherwise have as well.

```suggestion
      - name: Checkout
        uses: actions/checkout@v2
```

.github/workflows/docker_image.yaml
```diff
@@ -0,0 +1,39 @@
+name: Build and publish Docker Image
+
+on:
+  push:
+    tags:
+    - "*"
+
+jobs:
+  multi:
```
chaudum, line 9: I would name the job more specific, e.g. `docker`

chaudum (approved):

chaudum (approved):

MarkusH (approved):

Issue #101, opened 2020-11-06
Update finalizer, annotation storage and diff storage to use API_PREFIX


MarkusH:
[Kopf](https://kopf.readthedocs.io/) uses the prefix `kopf.zalando.org` in numerous places when handling custom resources.

* the `AnnotationsDiffBaseStorage` is initialized with `prefix='kopf.zalando.org'` and `key='last-handled-configuration'`
* the `AnnotationsProgressStorage` is initialized with `prefix='kopf.zalando.org'`
* the `PersistenceSettings.finalizer` is set to `'kopf.zalando.org/KopfFinalizerMarker'`

To better identify attributes written by the operator onto custom resources, I propose to change them to:

* `prefix='operator.cloud.crate.io'` and `key='last'` for `AnnotationsDiffBaseStorage`
* `prefix='operator.cloud.crate.io'` for `AnnotationsProgressStorage`
* `'operator.cloud.crate.io/finalizer'` for `PersistenceSettings.finalizer`

In order to cleanly transition between the existing storage configurations and the new ones one can use [`MultiDiffBaseStorage`](https://kopf.readthedocs.io/en/stable/packages/kopf.storage.diffbase/#kopf.storage.diffbase.MultiDiffBaseStorage) and [`MultiProgressStorage`](https://kopf.readthedocs.io/en/stable/packages/kopf.storage.progress/#kopf.storage.progress.MultiProgressStorage) respectively.

The general process to [transition storages is documented](https://kopf.readthedocs.io/en/stable/configuration/#storage-transition).

PR #107, opened 2020-11-11
Ref #101: Use built-in MultiDiffBaseStorage/MultiProgressStorage

MarkusH:

Ref #101
Partially revert 0e2bbd44a37c32d0c1e4eb5e613b2a7290c8ad79.

The previous change caused an issue in Kopf where it would no longer
detect changes, meaning sub-handlers would not re-execute. If a
sub-handler would fail, Kopf would patch the sub-handlers status into
the `progress_storage`. However, because that would also purge the
deprecated annotations, Kopf got confused in what happened and stopped
processing that resource.

To remedy the issue, the Kopf built-in `MultiDiffBaseStorage` and
`MultiProgressStorage` are now used. A future 2.0 release will be able
to remove the usage of the old annotations. Users are then able to
remove the old annotations from the custom resources themselves.

## Checklist


Taliik (approved):
lgtm

Issue #48, opened 2020-07-13
Improve handler idempotence


MarkusH:
## Problem Description

Kopf will invoke a handler repeatedly until it succeeds. That requires that all handlers are idempotent.

For the [create handler](https://github.com/crate/crate-operator/blob/60ce9b039ad26026a50fc840784e43ee4bf091fe/crate/operator/main.py#L113) we make use of the utility function [`call_kubeapi()`](https://github.com/crate/crate-operator/blob/60ce9b039ad26026a50fc840784e43ee4bf091fe/crate/operator/utils/kubeapi.py#L13) which optionally continues on missing or already present resources. For example, when creating a Kubernetes StatefulSet, the operator invokes `call_kubeapi()` with [`continue_on_conflict=True`](https://github.com/crate/crate-operator/blob/60ce9b039ad26026a50fc840784e43ee4bf091fe/crate/operator/create.py#L687-L690). This works, but doesn't follow Kopf's recommendation (see below).

For the [update handler](https://github.com/crate/crate-operator/blob/60ce9b039ad26026a50fc840784e43ee4bf091fe/crate/operator/main.py#L297) there's no real idempotence in play. Instead, some of the actions that handler takes will [retry themselves](https://github.com/crate/crate-operator/blob/60ce9b039ad26026a50fc840784e43ee4bf091fe/crate/operator/cratedb.py#L126) until they succeed or [run into some timeout](https://github.com/crate/crate-operator/blob/60ce9b039ad26026a50fc840784e43ee4bf091fe/crate/operator/main.py#L382-L389). This works for our current use but is less than ideal. It also doesn't follow Kopf's recommendation (see below).

## Kopf's Recommendation

In short, Kopf provides `kopf.register()` and `@kopf.on.this` as well as `kopf.execute()` to support idempotence. Details can be read up on https://kopf.readthedocs.io/en/stable/idempotence/. Essentially, these functions schedule sub-handlers which are guaranteed to only be run as often as necessary until they succeed. If they succeed in the first invocation, they won't be run again; if they fail for the first three times, and succeed on the fourth invocation, they are being called four time. That is, these invocations are _per invocation of the parent handler_.

For the creation, the change looks probably fairly straight forward. Instead invoking `call_kubeapi()`, the operator would move the creation of the sub-resources into their own sub-handlers.

[...]

PR #108, opened 2020-11-11
Ref #48: Make create handler more resilient

MarkusH:

Historically, the create handler for clusters was implemented in a
best-effort and with a bit of a "it'll be fine" approach. Kubernetes
resources were created and when they already existed that was fine. When
the create handler failed, though, the whole handler was being retried,
including the creation of already existing resources.

Now, with this change, only the individual sub-handlers will be retried,
making the process more efficient. The handler is now reentrant.

Unlike attempted in 07720e3080047c765211f6c56d70000b8b3a5043, the
ApiClient is now instantiated inside the individual sub-handlers, to
ensure a connection is available and open. Otherwise, sub-handlers will
fail with a `RuntimeError`: "Session is closed".

Refs #48
Refs 07720e3080047c765211f6c56d70000b8b3a5043

## Checklist

### DIFF
--- a/crate/operator/backup.py
+++ b/crate/operator/backup.py
@@ -15,7 +15,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import logging
-from typing import Any, Awaitable, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from kubernetes_asyncio.client import (
     AppsV1Api,
@@ -38,6 +38,7 @@ from kubernetes_asyncio.client import (
     V1PodTemplateSpec,
     V1SecretKeySelector,
 )
+from kubernetes_asyncio.client.api_client import ApiClient
 
 from crate.operator.config import config
 from crate.operator.constants import LABEL_COMPONENT, LABEL_NAME, SYSTEM_USERNAME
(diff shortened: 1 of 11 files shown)


crate/operator/backup.py
```diff
@@ -240,22 +241,24 @@ def create_backups(
...
-                    owner_references,
-                    name,
-                    labels,
-                    http_port,
-                    prometheus_port,
-                    backup_aws,
-                    image_pull_secrets,
-                    has_ssl,
+            )
+            deployment = (
```
chaudum, line 245: With this type of formatting it is not obvious that `deployment` will end up as type `tuple` - note the comma after the `call_kubeapi` call.

```
deployment = (
    await call_kubeapi(
        ...
    ),
)
```

I suggest to format like this:

```
deployment = await call_kubeapi(
...
)
```

I am wondering why mypy does not catch this case? :thinking:

chaudum (commented):
Otherwise lgtm. It's quite hard to review because of all the indentation changes

lukasbals (commented):
LGTM

crate/operator/backup.py
MarkusH: Wow. Nice catch!

chaudum (approved):

PR #114, opened 2020-11-18
Fix test configuration

lukasbals:

Fix test configuration

* Parse the path of the kubeconfig to allow relative paths (e.g. `~/.kube/kube_config`)
* Update kube context check to consider the `--kube-context` argument.

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,6 +16,7 @@
 
 import asyncio
 import os
+import pathlib
 import random
 import subprocess
 from unittest import mock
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -52,19 +53,29 @@ def pytest_configure(config):
...
-    if kubeconfig:
+    k8s_context = config.getoption(KUBECONTEXT_OPTION)
+
+    if kubeconfig and not k8s_context:
         p = subprocess.run(
-            ["kubectl", "--kubeconfig", kubeconfig, "config", "current-context"],
+            [
+                "kubectl",
+                "--kubeconfig",
+                str(pathlib.Path(kubeconfig).expanduser()),
```
MarkusH, line 63: Let's include an additional `resolve()` after `expanduser()` to get rid of relative path segments, such as `../`:
```suggestion
                str(pathlib.Path(kubeconfig).expanduser().resolve()),
```

tests/conftest.py
```diff
@@ -52,19 +53,29 @@ def pytest_configure(config):
...
         k8s_context = p.stdout.decode().strip()
 
-        if not k8s_context.startswith("crate-") and k8s_context != "minikube":
-            raise RuntimeError(
-                f"Cannot run tests on '{k8s_context}'. "
-                "Expected a context prefixed with 'crate-'."
-            )
+    if k8s_context and (
+        not k8s_context.startswith("crate-") and k8s_context != "minikube"
+    ):
```
MarkusH, line 74: I don't think the condition here is not quite is correct.

When no `kubeconfig` is specified, integration tests will not be run at all, thus no check on the `k8s_context` is necessary. However, if it is defined, the `k8s_context` can either come from the CLI argument or, if that's not specified, from the `kubectl config current-context` command. If the `k8s_context` then is either empty or doesn't start with `crate-` or `is `minikube`, then we should raise the exception:

```python
def ...():
    kubeconfig = config.getoption(KUBECONFIG_OPTION)
    if kubeconfig:
        k8s_context = config.getoption(KUBECONTEXT_OPTION)
        if k8s_context is None:
            k8s_context = subprocess.run(...)
        if k8s_context is None or not k8s_context.startswith("crate-") and k8s_context != "minikube":
            raise
```

lukasbals: As far as I tested it, the tests run if no `--kube-config` argument is passed. The other way to pass a config is by defining an env variable ( `KUBECONFIG`). As this is not documented in the development guide, I assume the default way for testing is the `--kube-config` argument?

So is it a bug that the tests are running if there is no `--kube-config` argument passed?

lukasbals: Good point 👍

MarkusH (approved):

PR #115, opened 2020-11-18
Fix #48: Improve resilience of update handler

### LINKED ISSUE #48: Improve handler idempotence

MarkusH:

This PR consists of three commits. The first one is mandatory, the second and third depend on the first.

Here's what the commits do:

1. Ref #48: Make update handler more resilient

   The operator used to have a long-running update handler. The handler takes care of updating, restarting, and scaling a cluster. However, due to the way the handler was implemented, it wasn't particular resilient. Operator restarts would require the whole handler to be executed again. This would cause all individual update tasks to be performed again as well. Especially with cluster restarts, that could cause a lot of problems, as nodes were restarted again, after having been restarted already.

   By using Kopf's sub-handler concept, individual sub-handlers are only executed once. However, in order to be able to order the sub-handler, a bit of state-handling is necessary. This commit implements the state-handling and allows for arbitrary sub-handlers to be run in an arbitrary order.

2. Ref #48: Allow cluster restart to be resumed

   Previously, if the operator would be restarted half-way through the rolling cluster restarts, the restart would start from the beginning. Especially on large clusters that could cause a lot of delay, if the operator would restart right before the last node was restarted.

   By raising `kopf.TemporaryError`s and by keeping track of the pods that need to be restarted, the restart operation can be resumed as often as necessary.

3. Ref #48: Allow cluster scale to be resumed

   Previously, if the operator would be restarted half-way through the scaling operation, the scaling would start from the beginning. Especially on large clusters that could cause a lot of delay.

   By raising `kopf.TemporaryError`s and checking for the state of nodes and replicas, the scaling operator can be resumed as often as necessary.

Fix #48


crate/operator/utils/kopf.py
```diff
@@ -26,3 +35,42 @@ async def _wrapper(**_):
...
+    def schedule_notification(self, event: WebhookEvent, payload: WebhookSubPayload):
+        self._context.schedule_notification(event, payload)
+
+    async def send_notifications(self, logger: logging.Logger):
+        for notification in self._context.notifications:
+            logger.info(
+                "Sending notification event %s with payload %s",
+                notification["event"],
+                notification["payload"],
+            )
```
lukasbals, line 72: It looks like the notifications are never sent 🤔

crate/operator/cratedb.py
```diff
@@ -109,11 +109,53 @@ async def get_healthiness(cursor: Cursor) -> int:
...
+        was shut down since the connection was opened.
+    :param expected_nodes: The number of nodes that make up a healthy cluster.
+    """
+    # We need to establish a new connection because the peer of a
+    # previous connection could have been shut down. And by
+    # re-establishing a connection for _each_ polling we can assert
+    # that the connection is open
+    async with connection_factory() as conn:
+        async with conn.cursor() as cursor:
+            logger.info("Waiting for cluster to get healthy again ...")
```
lukasbals, line 134: For me, it looks like here we are waiting to get the results of a health check and not waiting until a cluster is healthy. At least if you just consider the scope of the function. 🤔

crate/operator/main.py
```diff
@@ -382,129 +382,169 @@ async def cluster_create(
...
-async def cluster_update(
-    namespace: str,
-    name: str,
-    body: kopf.Body,
-    spec: kopf.Spec,
-    diff: kopf.Diff,
-    old: kopf.Body,
-    logger: logging.Logger,
-    **kwargs,
+@kopf.on.update(API_GROUP, "v1", RESOURCE_CRATEDB, id="update", backoff=10)
```
lukasbals, line 385: As discussed in the daily coffee, I'm not sure if the `id` and the `backoff` are needed.

crate/operator/main.py
```diff
@@ -382,129 +382,169 @@ async def cluster_create(
...
+        if new_cycle:
+            context.state_machine.add(State.SCALE)
+        kopf.register(
+            fn=ScaleSubHandler(context)(), id="scale", timeout=config.SCALING_TIMEOUT
+        )
+
+    patch.status["updateContext"] = context.serialize()
+
+
+class UpgradeSubHandler(StateBasedSubHandler):
```
lukasbals, line 433: I think moving those sub-handlers into a separate file (or separate files) would make sense to keep the main clean with just the handler invocations

crate/operator/operations.py
```diff
@@ -156,31 +122,73 @@ async def restart_cluster(
...
+                patch.status["pendingPods"] = pending_pods
+
+                raise kopf.TemporaryError(
+                    "Scheduling rerun because there are pods to be restarted", delay=5
+                )
+            else:
+                # We're all done
+                patch.status[
+                    "pendingPods"
+                ] = None  # Remove attribute from status stanca
```
lukasbals, line 184: What does `stance` mean? 🤔

crate/operator/main.py
MarkusH: Good idea!

crate/operator/cratedb.py
```diff
@@ -109,86 +108,43 @@ async def get_healthiness(cursor: Cursor) -> int:
...
+    # re-establishing a connection for _each_ polling we can assert
+    # that the connection is open
+    async with connection_factory() as conn:
+        async with conn.cursor() as cursor:
+            logger.info("Checking if cluster is healthy ...")
+            try:
+                num_nodes = await get_number_of_nodes(cursor)
+                healthiness = await get_healthiness(cursor)
+                if num_nodes == expected_nodes and healthiness in {1, None}:
+                    logger.info("Cluster has expected nodes and is healthy")
```
lukasbals, line 138:
```suggestion
                    logger.info("Cluster has expected number of nodes and is healthy")
```

crate/operator/main.py
```diff
@@ -382,191 +374,59 @@ async def cluster_create(
 
 
 @kopf.on.update(API_GROUP, "v1", RESOURCE_CRATEDB)
-async def cluster_update(
+async def update(
```
lukasbals, line 377: Wouldn't it make sense to have something in the name of this function that says that this function is for handling updates on custom CrateDB resources? 

Something like `update_crate_db` or `update_crate_db_resource`? 🤔

crate/operator/main.py
```diff
@@ -382,191 +374,59 @@ async def cluster_create(
...
+    if do_scale:
+        if new_cycle:
+            context.state_machine.add(State.SCALE)
+        kopf.register(
+            fn=ScaleSubHandler(namespace, name, context)(),
+            id="scale",
+            timeout=config.SCALING_TIMEOUT,
+        )
+
+    patch.status["updateContext"] = context.serialize()
```
lukasbals, line 429: This whole function looks very clean and well structured! Good work!

crate/operator/main.py
MarkusH: I guess `cluster_update` makes sense, doesn't it? That's then in sync with `cluster_create()` with `@kopf.on.create`.

crate/operator/main.py
MarkusH: :blush: Thank you!

crate/operator/main.py
lukasbals: Yeah, it also makes sense. 👍

lukasbals (approved):
As far as I can say, it looks good. 🥇

crate/operator/operations.py
```diff
@@ -156,31 +125,91 @@ async def restart_cluster(
...
+            delay=BACKOFF_TIME,
         )
-    for node_spec in cluster["spec"]["nodes"]["data"]:
-        await restart_statefulset(
-            core, conn_factory, namespace, name, node_spec["name"], total_nodes, logger
+    elif next_pod_name in all_pod_names:
+        total_nodes = get_total_nodes_count(old["spec"]["nodes"])
+        # The new pod has been spawned. Only a matter of time until it's ready.
+        password = await get_system_user_password(core, namespace, name)
+        host = await get_host(core, namespace, name)
```
chaudum, line 172: Could be done in parallel:

```suggestion
        password, host = await asyncio.gather(
            get_system_user_password(core, namespace, name),
            get_host(core, namespace, name),
        )
```

crate/operator/utils/kopf.py
```diff
@@ -26,3 +40,73 @@ async def _wrapper(**_):
...
+    #: retried using :class:`kopf.TemporaryError`.
+    state: State
+
+    def __init__(self, namespace: str, name: str, context: Context):
+        self.namespace = namespace
+        self.name = name
+        self._context = context
+
+    def __call__(self):
+        return functools.partial(self._subhandler)
```
chaudum, line 72: What difference does it make whether you return `self._subhandler` directly or as as partial without extra arguments?

crate/operator/operations.py
```diff
@@ -58,96 +53,70 @@ def get_total_nodes_count(nodes: Dict[str, Any]) -> int:
...
+) -> Tuple[Tuple[str], Tuple[str]]:
     """
-    Repeatedly and indefinitely check if ``pod``'s ``uid`` is still around.
+    Return a two-tuple with two tuples, the first containing all pod IDs, the
+    second the corresponding pod names within the cluster.
 
-    :param pod: The pod that is being terminated.
-    :param get_pods: A callable that returns an awaitable list of pods.
+    :param core: An instance of the Kubernetes Core V1 API.
+    :param namespace: The Kubernetes namespace where to look up CrateDB cluster.
```
chaudum, line 64:
```suggestion
    :param namespace: The Kubernetes namespace where to look up the CrateDB cluster.
```

crate/operator/main.py
```diff
@@ -386,188 +379,56 @@ async def cluster_create(
...
+            do_scale = True
 
-    async with ApiClient() as api_client:
-        apps = AppsV1Api(api_client)
-        coapi = CustomObjectsApi(api_client)
-        core = CoreV1Api(api_client)
+    if do_upgrade:
+        if new_cycle:
+            context.state_machine.add(State.UPGRADE)
+        kopf.register(fn=UpgradeSubHandler(namespace, name, context)(), id="upgrade")
```
chaudum, line 411: I find it a bit unconventional that you need to call the instantiated subhandler instance to get the subhandler function. I understand that it comes from `__call__` being a regular method, but the handler is an async method. The option to pass `UpgradeSubHandler(namespace, name, context)._subhandler` isn't pretty either.

chaudum (commented):
I really like the overall approach!

crate/operator/cratedb.py
```diff
@@ -125,86 +124,43 @@ async def get_healthiness(cursor: Cursor) -> int:
...
+    # We need to establish a new connection because the peer of a
+    # previous connection could have been shut down. And by
+    # re-establishing a connection for _each_ polling we can assert
+    # that the connection is open
+    async with connection_factory() as conn:
+        async with conn.cursor() as cursor:
+            logger.info("Checking if cluster is healthy ...")
+            try:
+                num_nodes = await get_number_of_nodes(cursor)
+                healthiness = await get_healthiness(cursor)
```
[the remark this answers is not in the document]
MarkusH: Good point. I think yes.

crate/operator/utils/kopf.py
MarkusH: I couldn't get it to work with `return self._subhandler`. or doing what's happening in `_subhandler` in `__call__` directly. The issues I faced were around pickling some thing in the inner parts of Kopf, where I then didn't spent time investigating. I _suspect_ it has something to do with how `kopf.register()` and async handlers / sub-handlers work. But yeah, I'm not too sure what exactly the underlying issue is.

crate/operator/main.py
MarkusH: As noted above, I originally planned on only passing the sub-handler instance and have a `async def __call_(self, **kwargs)` which does what `async def _subhandler(self, **kwargs)` does. However, that causes some errors inside serialization in Kopf which I didn't investigate further.

crate/operator/cratedb.py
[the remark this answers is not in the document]
MarkusH: Actually, not sure. Since they both use the same `cursor` object, wouldn't that potentially cause conflicts? Since both functions run `cursor.execute()` and `cursor.fetchone()`, couldn't the `fetchone()` in `get_healthiness()` return the response for the query in `get_number_of_nodes()`, if that's `execute()` was run later?

```
 get_number_of_nodes() | get_healthiness()   | What would cursor.fetchone() return
--------------------------------------------------------------------------------------------------
 cursor.execute(...)   |                     |
                       |                     | Result for number of nodes
                       | cursor.execute(...) |
                       |                     | Result for healthiness
 cursor.fetchone()     |                     | Returns the healthiness instead of number of nodes
                       | cursor.fetchone()   | Returns the healthiness instead of number of nodes


 get_number_of_nodes() | get_healthiness()   | What would cursor.fetchone() return
--------------------------------------------------------------------------------------------------
 cursor.execute(...)   |                     |
                       |                     | Result for number of nodes
 cursor.fetchone()     |                     | Returns the number of nodes
                       | cursor.execute(...) |
                       |                     | Result for healthiness
                       | cursor.fetchone()   | Returns the healthiness instead of number of nodes


 get_number_of_nodes() | get_healthiness()   | What would cursor.fetchone() return
--------------------------------------------------------------------------------------------------
                       | cursor.execute(...) |
                       |                     | Result for healthiness
 cursor.execute(...)   |                     |
                       |                     | Result for number of nodes
 cursor.fetchone()     |                     | Returns the number of nodes
                       | cursor.fetchone()   | Returns the number of nodes instead of healthiness
```

crate/operator/operations.py
```diff
@@ -156,31 +125,91 @@ async def restart_cluster(
...
-    password = await get_system_user_password(core, namespace, name)
-    host = await get_host(core, namespace, name)
-    conn_factory = connection_factory(host, password)
-
-    if "master" in cluster["spec"]["nodes"]:
-        await restart_statefulset(
-            core, conn_factory, namespace, name, "master", total_nodes, logger
+    pending_pods: List[Dict[str, str]] = status.get("pendingPods", [])
+    if not pending_pods:
+        pending_pods = []
```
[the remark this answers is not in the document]
MarkusH: Nice catch!

crate/operator/cratedb.py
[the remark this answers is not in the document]
chaudum: Right, you'd need separate cursor `objects`.

chaudum (approved):

PR #125, opened 2020-11-24
Allow overriding node.attr.zone

SStorm:

Before this change, even if you specified a node.attr.zone parameter it would have been explicitly overridden by the operator.

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,11 @@ Changelog
 Unreleased
 ----------
 
+* Fixed an inconsistent behavior where the configuration option
+  :envvar:`CLOUD_PROVIDER` would override an explicitly defined
+  ``node.attr.zone`` in either ``.spec.cluster.settings``,
+  ``.spec.nodes.master.settings``, or ``.spec.nodes.data.*.settings``.
+
 * To allow CrateDB user password updates, Kubernetes Secrets referenced in the
   ``.spec.users`` section of a CrateDB custom resource, will have a label
   ``operator.cloud.crate.io/user-password`` applied.
--- a/crate/operator/create.py
+++ b/crate/operator/create.py
@@ -385,6 +385,13 @@ def get_statefulset_crate_command(
             }
         )
 
+    if config.CLOUD_PROVIDER == CloudProvider.AWS:
+        url = "http://169.254.169.254/latest/meta-data/placement/availability-zone"
+        settings["-Cnode.attr.zone"] = f"$(curl -s '{url}')"
+    elif config.CLOUD_PROVIDER == CloudProvider.AZURE:
+        url = "http://169.254.169.254/metadata/instance/compute/zone?api-version=2020-06-01&format=text"  # noqa
+        settings["-Cnode.attr.zone"] = f"$(curl -s '{url}' -H 'Metadata: true')"
+
     if cluster_settings:
         for k, v in cluster_settings.items():
             settings[f"-C{k}"] = v
@@ -392,13 +399,6 @@ def get_statefulset_crate_command(
     node_settings = node_spec.get("settings", {})
     for k, v in node_settings.items():
         settings[f"-C{k}"] = v
-
-    if config.CLOUD_PROVIDER == CloudProvider.AWS:
-        url = "http://169.254.169.254/latest/meta-data/placement/availability-zone"
-        settings["-Cnode.attr.zone"] = f"$(curl -s '{url}')"
-    elif config.CLOUD_PROVIDER == CloudProvider.AZURE:
-        url = "http://169.254.169.254/metadata/instance/compute/zone?api-version=2020-06-01&format=text"  # noqa
-        settings["-Cnode.attr.zone"] = f"$(curl -s '{url}' -H 'Metadata: true')"
 
     return ["/docker-entrypoint.sh", "crate"] + [
         f"{k}={v}" for k, v in settings.items()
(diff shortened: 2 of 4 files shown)


MarkusH (commented):
I like the change :+1: .

There should be a short changelog entry for it, though. And I think it would be worth mentioning something about the zone in [docs/source/configuration.rst](https://crate-operator.readthedocs.io/en/latest/configuration.html#envvar-CLOUD_PROVIDER)

SStorm:
> I like the change  .
> 
> There should be a short changelog entry for it, though. And I think it would be worth mentioning something about the zone in [docs/source/configuration.rst](https://crate-operator.readthedocs.io/en/latest/configuration.html#envvar-CLOUD_PROVIDER)

Thanks, still learning the ropes :) Let me know what you think about the docs.

crate/operator/create.py
```diff
@@ -389,17 +389,17 @@ def get_statefulset_crate_command(
         for k, v in cluster_settings.items():
             settings[f"-C{k}"] = v
 
-    node_settings = node_spec.get("settings", {})
-    for k, v in node_settings.items():
-        settings[f"-C{k}"] = v
-
     if config.CLOUD_PROVIDER == CloudProvider.AWS:
```
MarkusH, line 392: Actually, I think this block of conditions needs to be moved before the `if cluster_settings` above (line 388). Otherwise, setting `.spec.cluster.settings['node.attr.zone']` there would still be overridden.

tests/test_create.py
```diff
@@ -379,6 +379,7 @@ def test_node_and_cluster_settings_may_override(self):
                     "auth.host_based.enabled": "node-override",
                     "node.attr.node_setting": "node-override",
                     "node.attr.some_node_setting": "node",
+                    "node.attr.zone": "$(curl -s 'https://example.com')",
```
MarkusH, line 382: I don't think the change in this test is valuable. There's no override happening, since the operator is not run with any `CLOUD_PROVIDER` setting during tests.

CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Changelog
 Unreleased
 ----------
 
+* The parameter ``node.attr.zone`` can now be overridden if required.
```
MarkusH, line 8: Sounds good. I'd be a bit more specific, though. When the operator was run with the `CLOUD_PROVIDER` setting set, the setting was overridden by the operator, ignoring the cluster or node settings. So, maybe something along these lines:

```rst
* Fixed an inconsistent behavior where the configuration option
  :envvar:`CLOUD_PROVIDER` would override an explicitly defined
  ``node.attr.zone`` in either ``.spec.cluster.settings``,
  ``.spec.nodes.master.settings``, or ``.spec.nodes.data.*.settings``.
```

docs/source/configuration.rst
```diff
@@ -43,6 +43,18 @@ expected to use upper-case letters and must be prefixed with
...
+   however in some cases this might need to be overridden. This can be done with:
+
+   .. code-block:: yaml
+
+      kind: CrateDB
+      spec:
+        cluster:
+          settings:
+            node.attr.zone: "$(curl -s 'https://example.com')"
+
```
MarkusH, line 57: This sounds good already! Some suggestions/rewordings, though.

restructuredText files typically have a line length of 79 characters. Would be nice to align with that here as well, to keep the docs consistent.

```suggestion
   Under the hood, the operator will pass a ``zone`` attribute to all CrateDB
   nodes. This attribute can also be defined explicitly or override the one set
   by the operator. To do this on a cluster level, set
   ``.spec.cluster.settings``:

   .. code-block:: yaml

      kind: CrateDB
      spec:
        cluster:
          settings:
            node.attr.zone: "some-value"

   To set or override the attribute on a node type level, set it in
   ``.spec.nodes.master.settings`` or ``.spec.nodes.data.*.settings``.
```

tests/test_create.py
```diff
@@ -430,6 +432,29 @@ def test_zone_attr(self, provider, url):
...
+                namespace="some-namespace",
+                name="cluster1",
+                master_nodes=["node-0", "node-1", "node-2"],
+                total_nodes_count=3,
+                crate_node_name_prefix="node-",
+                cluster_name="my-cluster",
+                node_name="node",
+                node_spec={
+                    "resources": {"cpus": 1, "disk": {"count": 1}},
+                    "settings": {"node.attr.zone": "$(curl -s 'https://example.com')"},
```
MarkusH, line 449: This should probably be moved to the `cluster_settings` to account for the issue pointed out above, where a cluster setting would still be overridden by the config option. And I'd probably use a simpler string for the zone value, e.g. `z1` or just "value" :man_shrugging:

crate/operator/create.py
SStorm: Agreed, well spotted thanks.

tests/test_create.py
SStorm: We have 2 code paths where `node.attr.zone` gets special treatment, so it felt like a good idea to also test the case where it doesn't get any special treatment. But ok, removed it.

tests/test_create.py
SStorm: I figured it might be a good idea to have a separate test cases for node_spec and cluster_settings. Let me know if you disagree!

MarkusH: Sounds good.

MarkusH (approved):
Well done :clap: :+1: :tada: :confetti_ball: 

In order to get the PR merged, add the "ready-to-merge" on the PR. [Mergify](https://mergify.io/) will then take care of the rest.

image

PR #126, opened 2020-11-24
Watch on Kubernetes Secrets and update user passwords if needed

lukasbals:

Watch on Kubernetes Secrets that have the `operator.cloud.crate.io/user-password` label assigned and update the users of all CrateDB resources in the same namespace if the password changed.

Story: https://trello.com/c/s8qWFI5L/102-8-password-recovery-for-cloud-clusters
Planning: https://github.com/crate/cloud-api/issues/1488

## Checklist

### DIFF
--- a/crate/operator/main.py
+++ b/crate/operator/main.py
@@ -53,9 +53,10 @@ from crate.operator.create import (
 from crate.operator.kube_auth import login_via_kubernetes_asyncio
 from crate.operator.operations import get_total_nodes_count, restart_cluster
 from crate.operator.scale import scale_cluster
+from crate.operator.update_user_password import update_user_password
 from crate.operator.upgrade import upgrade_cluster
 from crate.operator.utils.kopf import subhandler_partial
-from crate.operator.utils.kubeapi import ensure_user_password_label
+from crate.operator.utils.kubeapi import ensure_user_password_label, get_host
 from crate.operator.webhooks import (
     WebhookScaleNodePayload,
     WebhookScalePayload,
@@ -592,3 +593,51 @@ async def update_cratedb_resource(
                     await ensure_user_password_label(
                         core, namespace, user_spec["password"]["secretKeyRef"]["name"]
                     )
+
+
+@kopf.on.update("", "v1", "secrets", labels={LABEL_USER_PASSWORD: "true"})
+async def secret_update(
+    namespace: str,
+    name: str,
+    diff: kopf.Diff,
+    logger: logging.Logger,
+    **kwargs,
+):
+    async with ApiClient() as api_client:
+        coapi = CustomObjectsApi(api_client)
+        core = CoreV1Api(api_client)
+
+        for operation, field_path, old_value, new_value in diff:
+            custom_objects = await coapi.list_namespaced_custom_object(
+                namespace=namespace,
+                group=API_GROUP,
+                version="v1",
+                plural=RESOURCE_CRATEDB,
+            )
+
+            for crate_custom_object in custom_objects["items"]:
+                host = await get_host(
+                    core, namespace, crate_custom_object["metadata"]["name"]
+                )
+
+                for user_spec in crate_custom_object["spec"]["users"]:
+                    expected_field_path = (
+                        "data",
+                        user_spec["password"]["secretKeyRef"]["key"],
+                    )
+                    if (
+                        user_spec["password"]["secretKeyRef"]["name"] == name
+                        and field_path == expected_field_path
+                    ):
+                        kopf.register(
+                            fn=subhandler_partial(
+                                update_user_password,
+                                host,
+                                user_spec["name"],
+                                old_value,
+                                new_value,
+                                logger,
+                            ),
+                            id=f"update-{crate_custom_object['metadata']['name']}-{user_spec['name']}",  # noqa
+                            timeout=config.BOOTSTRAP_TIMEOUT,
+                        )
--- a/crate/operator/update_user_password.py
+++ b/crate/operator/update_user_password.py
@@ -1,0 +1,46 @@
+# CrateDB Kubernetes Operator
+# Copyright (C) 2020 Crate.IO GmbH
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import logging
+
+from crate.operator.constants import BACKOFF_TIME
+from crate.operator.cratedb import get_connection, update_user
+from crate.operator.utils.formatting import b64decode
+
+
+# update_user_password(host, username, old_password, new_password)
+async def update_user_password(
+    host: str,
+    username: str,
+    old_password: str,
+    new_password: str,
+    logger: logging.Logger,
+):
+    """
+    Update the password of a given ``user_spec`` in a CrateDB cluster.
+
+    :param host: The host of the CrateDB resource that should be updated.
+    :param username: The username of the user of the CrateDB resource that
+        should be updated.
+    :param old_password: The old password of the user that should be updated.
+    :param new_password: The new password of the user that should be updated.
+    """
+    async with get_connection(
+        host, b64decode(old_password), username, timeout=BACKOFF_TIME / 4.0
+    ) as conn:
+        async with conn.cursor() as cursor:
+            logger.info("Updating password for user '%s'", username)
+            await update_user(cursor, username, b64decode(new_password))
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -33,6 +33,27 @@ The entire bootstrapping process may not take longer than (by default) 1800
 seconds before it is considered failed. The timeout can be configured with the
 :envvar:`BOOTSTRAP_TIMEOUT` environment variable.
 
+User Passwords
+~~~~~~~~~~~~~~
+
+When creating the CrateDB users specified under ``.spec.users``, the operator
+will add the ``operator.cloud.crate.io/user-password`` label to each of the
+Kubernetes Secrets assigned to one of the users. To keep backward
+compatibility, it also adds the label to Kubernetes Secrets referenced in
+existing CrateDB resources :func:`on resume <kopf:kopf.on.resume>`.
+
+The ``operator.cloud.crate.io/user-password`` label is used to filter the
+events when watching for changes on one of the Kubernetes Secrets. If one of
+the Kubernetes Secrets is updated, the operator will update all CrateDB users
+that use that secret by iterating over all CrateDB resources. The operator
+updates the password in a CrateDB cluster by logging in to one of the CrateDB
+nodes with the corresponding username and old password. It will then use the
+:ref:`cratedb:ref-alter-user` query to update the password.
+
+.. note::
+
+   If one changes the CrateDB user's password directly in CrateDB, the operator
+   won't be able to update that user anymore.
 
 Cluster Restart
 ---------------
--- a/tests/test_update_password.py
+++ b/tests/test_update_password.py
@@ -1,0 +1,148 @@
+# CrateDB Kubernetes Operator
+# Copyright (C) 2020 Crate.IO GmbH
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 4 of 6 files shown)


crate/operator/main.py
```diff
@@ -592,3 +593,41 @@ async def update_cratedb_resource(
...
+                for crate_custom_object in custom_objects["items"]:
+                    crate_custom_object_users = crate_custom_object["spec"]["users"]
+
+                    updated_user_spec = None
+                    for user_spec in crate_custom_object_users:
+                        if name == user_spec["password"]["secretKeyRef"]["name"]:
+                            updated_user_spec = user_spec
+
+                    if updated_user_spec is not None:
+                        await update_user_password(
```
lukasbals, line 627: Note to myself: Check if the same secret can be used in the same CrateDB resource for multiple users.

lukasbals: I think a secret **can** be used in multiple users of one CrateDB resource. Therefore, I updated the logic.

crate/operator/main.py
```diff
@@ -592,3 +593,37 @@ async def update_cratedb_resource(
...
+
+@kopf.on.update("", "v1", "secrets", labels={LABEL_USER_PASSWORD: "true"})
+async def secret_update(
+    namespace: str,
+    name: str,
+    diff: kopf.Diff,
+    **kwargs,
+):
+    for operation, field_path, old_value, new_value in diff:
+        if field_path == ("data", "password"):
```
MarkusH, line 606: `password` isn't guaranteed to be the key in the secret that contains a password. It can be anything. The password that is used for a specific user is defined in the custom resource under `.spec.users.*.password.secretKeyRef.key`.

In essence, I think you'll need to iterate over all users in all cluster custom resources in the namespace of this secret, and check if the password for that user was changed:

```
for cluster in all_clusters_in_namespace:
    for user_spec in cluster.users:
        if user_spec.secretKeyRef.name == name_of_secret_that_was_changed and user_spec.secretKeyRef.key in diff:
            update_user_password
```

crate/operator/update_user_password.py
```diff
@@ -0,0 +1,54 @@
...
+
+    :param core: An instance of the Kubernetes Core V1 API.
+    :param namespace: The Kubernetes namespace for the CrateDB cluster.
+    :param name: The name for the ``CrateDB`` custom resource. Used to lookup
+        the password for the system user created during deployment.
+    :param user_spec: The user specification from the ``CrateDB`` custom resource
+        users.
+    :param old_password: The old password of the user that should be updated.
+    """
+    host = await get_host(core, namespace, name)
```
MarkusH, line 45: The way you're calling `update_user_password()`, `get_host()` might be called numerous times for each cluster.

I think a function signature `update_user_password(host, username, old_password, new_password) should be sufficient. You can call `get_host()` once in the `secret_update` handler, and you already have the old and new password in the diff that you're iterating over. You only need to decode both from base64 to plain text.

crate/operator/main.py
```diff
@@ -592,3 +593,37 @@ async def update_cratedb_resource(
...
+                    version="v1",
+                    plural=RESOURCE_CRATEDB,
+                )
+
+                for crate_custom_object in custom_objects["items"]:
+                    crate_custom_object_users = crate_custom_object["spec"]["users"]
+
+                    for user_spec in crate_custom_object_users:
+                        if name == user_spec["password"]["secretKeyRef"]["name"]:
+                            await update_user_password(
```
MarkusH, line 623: Instead of awaiting each password update, which essentially means it's all synchronously, I think using something along the lines of what we do with the statefulsets in the create handler would make sense: https://github.com/crate/crate-operator/blob/0cc2d582976b2a21e69674c99b2c958233a07a2d/crate/operator/main.py#L309-L338

```python
kopf.register(
    fn=subhandler_partial(
        update_user_password,
        host,
        username,
        old_password,
        new_password,
    ),
    id=f"update-{crate_custom_object.metadata.name}-{user_spec.name}",
    timeout=config.BOOTSTRAP_TIMEOUT,  # Not sure about this, but the operator should give up _at some point_.
)

File: crate/operator/main.py (hunk 1 above)
Comment (MarkusH) [reply crate/operator/main.py]: Yes, it can.

File: crate/operator/main.py (hunk 2 above)
Comment (lukasbals) [reply crate/operator/main.py:606]: It makes sense ... I haven't been sure about that.

File: crate/operator/update_user_password.py (hunk 3 above)
Comment (lukasbals) [reply crate/operator/update_user_password.py:45]: Good point ... makes sense.

The only downside I see is that we have to instantiate `CoreV1Api` before we even know if there is something we have to update. But yes, we save a lot more time if we await the host just once.

Comment (MarkusH) [reply crate/operator/update_user_password.py:45]: Yeah, but instantiating `CoreV1Api` is much easier than querying for the host ;-)

File: crate/operator/main.py [hunk 5]
```diff
@@ -592,3 +593,45 @@ async def update_cratedb_resource(
...
+            for crate_custom_object in custom_objects["items"]:
+                host = await get_host(
+                    core, namespace, crate_custom_object["metadata"]["name"]
+                )
+
+                for user_spec in crate_custom_object["spec"]["users"]:
+                    if (
+                        user_spec["password"]["secretKeyRef"]["name"] == name
+                        and user_spec["password"]["secretKeyRef"]["key"] in field_path
+                    ):
```
MarkusH, line 626: Regular K8s secrets' data is always in the `data` part of the resource. Which means, the password for a user is in `user_spec["password"]["secretKeyRef"]["key"]` underneath `data`. Hence, you can check for equality instead of using `in`

```suggestion
                    expected_field_path = ("data", user_spec["password"]["secretKeyRef"]["key"])
                    if (
                        user_spec["password"]["secretKeyRef"]["name"] == name
                        and field_path == expected_field_path
                    ):
```

tests/test_update_password.py
```diff
@@ -0,0 +1,170 @@
...
+        core, namespace.metadata.name, name
+    )
+    await assert_wait_for(
+        True,
+        is_password_set,
+        host,
+        system_password,
+        SYSTEM_USERNAME,
+        timeout=BACKOFF_TIME * 5,
+    )
```
MarkusH, line 144: I don't think those two checks are needed. The `is_password_set` right below (line 148) will only succeed if a system user was created before.

tests/test_update_password.py
```diff
@@ -0,0 +1,170 @@
...
+        ),
+    )
+
+    await assert_wait_for(
+        True,
+        is_password_set,
+        host,
+        new_password,
+        username,
+        timeout=BACKOFF_TIME * 5,
```
MarkusH, line 169: This should not take more than a minute or two, IMO. There's shouldn't be any need for retry or whatnot.

crate/operator/update_user_password.py
```diff
@@ -0,0 +1,42 @@
...
+    :param username: The username of the user of the CrateDB resource that
+        should be updated.
+    :param old_password: The old password of the user that should be updated.
+    :param new_password: The new password of the user that should be updated.
+    """
+    async with get_connection(
+        host, b64decode(old_password), username, timeout=BACKOFF_TIME / 4.0
+    ) as conn:
+        async with conn.cursor() as cursor:
+            await update_user(cursor, username, b64decode(new_password))
```
MarkusH, line 42: Let's emit a log message here.
```python
async def update_user_password(..., logger: logging.Logger):
    ...
    logger.info("Updating password for user '%s'", username)
    await update_user(cursor, username, b64decode(new_password))
```

MarkusH (commented):
:clap: 

This change should also include updates to the documentation. I could imagine a sub-paragraph under https://crate-operator.readthedocs.io/en/latest/concepts.html#bootstrapping . @infoverload ? Essentially, what I'd love to have in there is how the upgrade works. Like,

* the operator adding the labels to the secrets to track them during `create` and `resume`
* the operator logging in to CrateDB with the user's old secrets and then updating them
* that the operator won't be able to upgrade passwords that have been changed directly through CrateDB and not through the operator.

MarkusH:
While writing https://github.com/crate/crate-operator/pull/126#pullrequestreview-538451081, what we haven't talked about it, how the operator deals with those secrets when clusters are deleted. Maybe the operator should go through all secrets that are used by the deleted cluster, check if they are used by another cluster, and if not, delete the label?

tests/test_update_password.py
lukasbals: I just added them to make the test easier to understand ... but I can, of course, remove it.

MarkusH: Na, it's fine :)

lukasbals:
> While writing [#126 (review)](https://github.com/crate/crate-operator/pull/126#pullrequestreview-538451081), what we haven't talked about it, how the operator deals with those secrets when clusters are deleted. Maybe the operator should go through all secrets that are used by the deleted cluster, check if they are used by another cluster, and if not, delete the label?

The purpose would be to clean up the secrets and maybe use them for something else, right?
I'm not sure if it makes sense to remove the label from the secret, to be honest. 

Anyway, if we do implement that functionality to remove the label from the secrets on the deletion of the cluster I would do it in a new PR. What do you think?

MarkusH:
> > While writing [#126 (review)](https://github.com/crate/crate-operator/pull/126#pullrequestreview-538451081), what we haven't talked about it, how the operator deals with those secrets when clusters are deleted. Maybe the operator should go through all secrets that are used by the deleted cluster, check if they are used by another cluster, and if not, delete the label?
> 
> The purpose would be to clean up the secrets and maybe use them for something else, right?
> I'm not sure if it makes sense to remove the label from the secret, to be honest.

It was just a thought. Essentially, the operator should take care of all the cleanup as well.

> Anyway, if we do implement that functionality to remove the label from the secrets on the deletion of the cluster I would do it in a new PR. What do you think?

Yes, absolutely. A separate PR definitely makes sense.

docs/source/concepts.rst
```diff
@@ -33,6 +33,24 @@ The entire bootstrapping process may not take longer than (by default) 1800
...
+events when watching for changes on one of the Kubernetes Secrets. If one of
+the Kubernetes Secrets with the ``operator.cloud.crate.io/user-password``-label
+assigned is updated, the Operator will update all CrateDB users that use that
+secret by iterating over all CrateDB resources. The Operator does the update of
+the password in a CrateDB cluster by logging in to one of the master nodes of
+the CrateDB cluster using the username defined in the CrateDB custom resource
+definition and the old password.
+
+Please note that if one updates the CrateDB user directly on one of the master
+nodes, the Operator won't be able to update that user anymore.
```
MarkusH, line 53: A bit of rewording:

* As far as I can tell, the K8s convention is to start with a `.`.
* "Operator" is not a name and thus not spelled with a capital "O". Unless used in "CrateDB Kubernetes Operator".
* You can use something called [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) to link between Sphinx configurations. The setup for that is in the Sphinx configuration: https://github.com/crate/crate-operator/blob/0cc2d582976b2a21e69674c99b2c958233a07a2d/docs/source/conf.py#L75-L82 . To find the link targets in the target documentation, you can use `python -msphinx.ext.intersphinx URL`, e.g. `python -msphinx.ext.intersphinx https://crate.io/docs/crate/reference/en/stable/objects.inv`. Leveraging intersphinx will help against dead links and adds contextual information to Sphinx what you're linking to. E.g. the "on resume" link looks differently that the one for "ALTER USER", because one is a function (`:func:`) and the other is just a regular reference (`:ref:`). And yeah, I notice, there are other links that should probably use intersphinx.

```suggestion
User Password
~~~~~~~~~~~~~

When creating the CrateDB users specified under ``.spec.users``, the operator
will add the ``operator.cloud.crate.io/user-password`` label to each of the
Kubernetes Secrets assigned to one of the users. To keep backward
compatibility, it also adds the label to Kubernetes Secrets referenced in
existing CrateDB resources :func:`on resume <kopf:kopf.on.resume>`.

The ``operator.cloud.crate.io/user-password`` label is used to filter the
events when watching for changes on one of the Kubernetes Secrets. If one of
the Kubernetes Secrets is updated, the operator will update all CrateDB users
that use that secret by iterating over all CrateDB resources. The operator does
the update of the password in a CrateDB cluster by logging in to one of the
CrateDB nodes with the corresponding username and old password. It will then
use the :ref:`cratedb:ref-alter-user` query to update the password.

.. note::

   If one changes the CrateDB user's password directly in CrateDB, the operator
   won't be able to update that user anymore.
```

lukasbals: Thanks for all the explanation!  Intersphinx looks really cool ... never heard about it before.

MarkusH (approved):
Looks great :tada: . I would love from somebody from @crate/tech-writing to have a final look at the docs update before merging. Maybe @infoverload can spare a few minutes to check with you?

docs/source/concepts.rst
```diff
@@ -33,6 +33,27 @@ The entire bootstrapping process may not take longer than (by default) 1800
...
+will add the ``operator.cloud.crate.io/user-password`` label to each of the
+Kubernetes Secrets assigned to one of the users. To keep backward
+compatibility, it also adds the label to Kubernetes Secrets referenced in
+existing CrateDB resources :func:`on resume <kopf:kopf.on.resume>`.
+
+The ``operator.cloud.crate.io/user-password`` label is used to filter the
+events when watching for changes on one of the Kubernetes Secrets. If one of
+the Kubernetes Secrets is updated, the operator will update all CrateDB users
+that use that secret by iterating over all CrateDB resources. The operator does
+the update of the password in a CrateDB cluster by logging in to one of the
```
infoverload, line 49: I would just write "updates" instead of "does the update of"

PR #133, opened 2020-11-30
Update permissions of the service account to be able to watch secrets

lukasbals:

Update permissions of the service account to be able to watch secrets.

## Checklist

### DIFF
--- a/deploy/rbac.yaml
+++ b/deploy/rbac.yaml
@@ -68,6 +68,7 @@ rules:
   - list
   - patch
   - update
+  - watch
 - apiGroups:
   - ""
   resources:


CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Changelog
 Unreleased
 ----------
 
+* Updated permissions of the service account to be able to watch secrets.
+
```
chaudum, line 9: I don't think you need this changelog entry, since the change is part of https://github.com/crate/crate-operator/commit/a19d9447bec053b37c9deb35966b7f1e40b93dcc (which has not been released yet).

chaudum (commented):
Otherwise ok!

CHANGES.rst
lukasbals: Good point

chaudum (approved):

PR #134, opened 2020-11-30
Pass the webhook params to the backup job

SStorm:

We want the backup jobs to notify us when they are complete (or failed). For this, pass on the webhook configuration that is already present in the operator to the created cronjob.

The params are passed as plain text variables since there's no obvious benefit in passing them as secrets (and it's much harder to do in a generic way).

## Checklist

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -64,7 +64,9 @@ expected to use upper-case letters and must be prefixed with
 
    When enabling backups for a cluster, the operator deploys a Prometheus_
    exporter to be scraped for backup metrics, and a Kubernetes CronJob that
-   creates backups every defined interval. This variable needs to point to a
+   creates backups every defined interval. If :envvar:`WEBHOOK_URL` and
+   related credentials are specified, the backup CronJob will post backup
+   creation events back to the webhook URL. This variable needs to point to a
    Docker image *and* tag to use it for the exporter and CronJob.
 
 .. envvar:: DEBUG_VOLUME_SIZE
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changelog
 Unreleased
 ----------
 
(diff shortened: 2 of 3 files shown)


crate/operator/config.py
```diff
@@ -82,16 +82,16 @@ class Config:
     #: allow for easier testing in smaller Kubernetes clusters.
     TESTING: bool = False
 
-    #: HTTP Basic Auth password for web requests made to :attr:`WEBHOOK_URL`.
```
MarkusH, line 85: The settings are deliberately sorted this way: alphabetical 😉. Even though username and password belong together, "password" comes before "URL" . Same for the assignment below inside `load()`.

SStorm: Who knew! :) Thanks, I'll revert.

chaudum (commented):
I think it would make sense to add some words about passing these environment variables into the deployment of the backup container.

https://github.com/crate/crate-operator/blame/master/docs/source/configuration.rst#L65-L68

Please also add a changelog entry to CHANGES.rst

docs/source/configuration.rst
```diff
@@ -64,7 +64,9 @@ expected to use upper-case letters and must be prefixed with
 
    When enabling backups for a cluster, the operator deploys a Prometheus_
    exporter to be scraped for backup metrics, and a Kubernetes CronJob that
-   creates backups every defined interval. This variable needs to point to a
+   creates backups every defined interval. If :envvar:`WEBHOOK_URL` and
+   related credentials are specified, the backup CronJob will post backup
+   creation events back to the Cloud API. This variable needs to point to a
```
chaudum, line 69:
```suggestion
   creation events back to the webhook URL. This variable needs to point to a
```

chaudum (commented):
Please replace the reference to the Cloud API in the docs. Otherwise lgtm!

chaudum (approved):

chaudum (approved):

PR #137, opened 2020-12-04
Use default backoff time when raising TemporaryError

chaudum:

The [default value for backoff](https://github.com/nolar/kopf/blob/a6d302bbf1efc39e371167ed66dbbfadff8a66ca/kopf/reactor/handling.py#L21) is defined in Kopf as `1min`, which is the
same as the value we provided.
Therefore we can get rid of the `BACKOFF_TIME` constant and instead use
dedicated a `CONNECT_TIMEOUT` when creating new database connections and
a `DEFAULT_TIMEOUT` for tests.

## Checklist

### DIFF
--- a/crate/operator/constants.py
+++ b/crate/operator/constants.py
@@ -28,7 +28,7 @@ LABEL_USER_PASSWORD = f"operator.{API_GROUP}/user-password"
 
 SYSTEM_USERNAME = "system"
 
-BACKOFF_TIME = 60.0
+CONNECT_TIMEOUT = 10.0
 
 
 class CloudProvider(str, enum.Enum):
--- a/crate/operator/utils/kubeapi.py
+++ b/crate/operator/utils/kubeapi.py
@@ -21,7 +21,7 @@ from typing import Awaitable, Callable, Optional
 from kubernetes_asyncio.client import ApiException, CoreV1Api, V1ObjectMeta, V1Secret
 
 from crate.operator.config import config
-from crate.operator.constants import BACKOFF_TIME, LABEL_USER_PASSWORD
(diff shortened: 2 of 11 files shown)


crate/operator/constants.py
```diff
@@ -28,7 +28,7 @@
 
 SYSTEM_USERNAME = "system"
 
-BACKOFF_TIME = 60.0
+CONNECT_TIMEOUT = 10.0
```
SStorm, line 31: Aren't these somewhat different concepts? `CONNECT_TIMEOUT` would suggest to me that it's simply a "fail if not connected by this time" rather than a "delay reconnect by this time"?

crate/operator/utils/kubeapi.py
```diff
@@ -162,7 +162,7 @@ async def get_public_host(core: CoreV1Api, namespace: str, name: str) -> str:
         except ApiException:
             pass
 
-        await asyncio.sleep(BACKOFF_TIME / 2)
+        await asyncio.sleep(10)
```
SStorm, line 165: Maybe we can set this to something even lower? Doesn't cost us much to make this API call to kubernetes when retrying...

chaudum: Indeed. What about 5s?

crate/operator/constants.py
chaudum: Right, that's why I changed the name of the constant to CONNECT_TIMEOUT, and it is only used as argument of the `get_connection()` method https://github.com/crate/crate-operator/pull/137/files#diff-1af5aaa975261c7031681203924cb7afde32559b890a02d25a43cbe4531d8c93R283

SStorm (approved):

PR #138, opened 2020-12-07
Extend docs on how to use Minikube

chaudum:

Without the correct storage class, the tests would run into timeouts.

## Checklist

### DIFF
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -87,15 +87,45 @@ passed via the ``--kube-config`` and ``--kube-context`` arguments to pytest_.
 Furthermore, the context *must* start with either ``crate-`` or be called
 ``minikube``.
 
-.. note::
+Minikube
+^^^^^^^^
 
-   If you are using macOS in combination with Minikube, you need to set up a
-   `Minikube tunnel`_ in order to get an external IP address assigned to the
-   CrateDB services.
+Using minikube_ as Kubernetes backend for the tests requires a few
+preparations. For example, make sure your minikube has configured enough
+resources for example:
 
-   .. code-block:: console
+.. code-block:: console
 
-      $ minikube tunnel
+   $ minikube config view
+   - cpus: 4
+   - disk-size: 64G
+   - memory: 8192
+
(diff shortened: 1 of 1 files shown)


docs/source/development.rst
```diff
@@ -87,23 +87,54 @@ passed via the ``--kube-config`` and ``--kube-context`` arguments to pytest_.
...
 
-.. note::
+Minikube
+........
 
-   If you are using macOS in combination with Minikube, you need to set up a
-   `Minikube tunnel`_ in order to get an external IP address assigned to the
-   CrateDB services.
+Using minikube_ as Kubernetes backend for the tests requires a few
+preparations. Make sure, your minikube has configured enough resources, for
```
SStorm, line 94: The commas are not necessary here?

```suggestion
preparations. Make sure your minikube has configured enough resources for
```

SStorm (approved):
One minor comment, but looks good!

docs/source/development.rst
chaudum: No, don't think so :+1:

docs/source/development.rst
```diff
@@ -87,23 +87,54 @@ passed via the ``--kube-config`` and ``--kube-context`` arguments to pytest_.
...
 
-.. note::
+Minikube
+^^^^^^^^
 
-   If you are using macOS in combination with Minikube, you need to set up a
-   `Minikube tunnel`_ in order to get an external IP address assigned to the
-   CrateDB services.
+Using minikube_ as Kubernetes backend for the tests requires a few
+preparations. Make sure your minikube has configured enough resources for
```
matthijskrul, line 94: A little more natural is "For example, make sure your Minikube...". Also think about whether you want to capitalize Minikube or not - either is fine but should be consistent

chaudum: I think lower case is correct. https://minikube.sigs.k8s.io/docs/

chaudum: If it's lower case, do you mean it should also be lower case in the title?

PR #141, opened 2020-12-09
Added max-shards-per-node metric to the sql exporter

juanpardo:
## Checklist

### DIFF
--- a/crate/operator/create.py
+++ b/crate/operator/create.py
@@ -21,6 +21,7 @@ import warnings
 from typing import Any, Dict, List, Optional, Tuple
 
 import bitmath
+import yaml
 from kubernetes_asyncio.client import (
     AppsV1Api,
     CoreV1Api,
@@ -155,12 +156,15 @@ async def create_debug_volume(
 def get_sql_exporter_config(
     owner_references: Optional[List[V1OwnerReference]], name: str, labels: LabelType
 ) -> V1ConfigMap:
+
     sql_exporter_config = pkgutil.get_data("crate.operator", "data/sql-exporter.yaml")
-    responsivity_collector_config = pkgutil.get_data(
-        "crate.operator", "data/responsivity-collector.yaml"
-    )
-    if sql_exporter_config and responsivity_collector_config:
-        return V1ConfigMap(
+
+    if sql_exporter_config:
+        # Parse the config yaml file to get the defined collectors and load them
+        parsed_sql_exporter_config = yaml.load(sql_exporter_config.decode())
+        collectors = parsed_sql_exporter_config["target"]["collectors"]
+
+        result = V1ConfigMap(
             metadata=V1ObjectMeta(
                 name=f"crate-sql-exporter-{name}",
                 labels=labels,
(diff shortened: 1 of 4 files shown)


juanpardo:
Tested locally against CrateDB 4.3.2, got the following output:
```# HELP max_shards_per_node Indicates the maximum number of shards allowed per node defined in the configuration of the cluster.
# TYPE max_shards_per_node gauge
max_shards_per_node{max_shards_per_node="max_shards_per_node"} 1000
# HELP responsivity Indicates whether the CrateDB node is responding to queries. Will not return if the node is stuck.
# TYPE responsivity gauge
responsivity{responsive="responsive"} 1
responsivity{responsive="states"} 0```

**Comment by chaudum** [comment]
> Tested locally against CrateDB 4.3.2, got the following output:
> 
> ```
> # HELP max_shards_per_node Indicates the maximum number of shards allowed per node defined in the configuration of the cluster.
> # TYPE max_shards_per_node gauge
> max_shards_per_node{max_shards_per_node="max_shards_per_node"} 1000
> # HELP responsivity Indicates whether the CrateDB node is responding to queries. Will not return if the node is stuck.
> # TYPE responsivity gauge
> responsivity{responsive="responsive"} 1
> responsivity{responsive="states"} 0
> ```

I think we should follow the [best practices for naming of metrics and labels](https://prometheus.io/docs/practices/naming/) so I would suggest to name the metric `cratedb_max_shards_per_node`. I also think that the label does not really make sense, especially in combination with its value. Can we omit it? We would end up with the metric `cratedb_max_shards_per_node{} 1000`.

/cc @christianbader

**Comment by turbo-ele** [comment]
I agree with the comments made by @chaudum. Actually, also `responsivity` should therefore be called `cratedb_responsivity`, but this is out of scope for this task I'd say.

File: crate/operator/create.py [hunk 1]
```diff
@@ -155,22 +156,41 @@ async def create_debug_volume(
...
             data={
                 "sql-exporter.yaml": sql_exporter_config.decode(),
-                "responsivity-collector.yaml": responsivity_collector_config.decode(),
             },
         )
+
+        # Add the yaml collectors to the configmap dynamically
+        for collector in collectors:
+            if collector.endswith("_collector"):
+                collector = collector[:-10]
```
turbo-ele, line 181: alternatively, for better readability
```suggestion
                collector = collector.replace("_collector", "")
```

crate/operator/data/cratedb_max_shard-collector.yaml
```diff
@@ -0,0 +1,17 @@
+---
+collector_name: cratedb_max_shards_collector
+# This metric only publishes the configuration set at the cluster level of the maximum number of shards a node can
+# handle.
+# This metric is to be used to alert us when the current number of shards per node surpasses a certain threshold.
+# Notice the calculation happens at the Prometheus level and not here so that the metrics can be kept generic
```
turbo-ele, line 6:
```suggestion
# Notice the calculation happens at Prometheus level and not here so that the metrics can be kept generic
```

tests/test_create.py
```diff
@@ -941,3 +942,17 @@ async def test_preserve_unknown_object_keys(
         assert resource["spec"]["nodes"]["master"]["labels"] == {"s.n.m.l": "1"}
         assert resource["spec"]["nodes"]["master"]["settings"] == {"s.n.m.s": "1"}
         assert resource["status"] == {"foo": "bar", "buz": {"lorem": "ipsum"}}
+
+
+def test_sql_exporter_config():
+    result = get_sql_exporter_config(None, "test-name", None)
+    assert result.metadata.name == "crate-sql-exporter-test-name"
+
+    from os import walk
```
turbo-ele, line 951: import on top of the file

turbo-ele (approved):
lgtm! 👍

crate/operator/create.py
juanpardo: I would rather add a comment, since the proposed change does not guarantee it will delete only the suffix. I'm probably being too picky on this one though!

tests/test_create.py
juanpardo: isort 5 does not complain about this anymore by default!

PR #155, opened 2021-01-25
Remove username validation from the custom resource definition

lukasbals:
#### Remove username validation from the custom resource definition.

Since CrateDB accepts every string as a username we also don't want
that check in the crate-operator.

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,10 @@
 Changelog
 =========
 
+* Removed username validation from the custom resource definition.
+  Since CrateDB accepts every string as a username, we also don't want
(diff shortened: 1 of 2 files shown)


SStorm (approved):

CHANGES.rst
```diff
@@ -2,6 +2,10 @@
 Changelog
 =========
 
+* Removed username validation from the custom resource definition.
+  Since CrateDB accepts every string as a username we also don't want
```
turbo-ele, line 6:
```suggestion
  Since CrateDB accepts every string as a username, we also don't want
```

turbo-ele (approved):

PR #160, opened 2021-01-26
Revert local StorageClass

SStorm:

This reverts the change that introduced the local StorageClass. The SC seems to be causing issues in Azure, and the reasoning why it was introduced is unclear.

## Checklist

### DIFF
--- a/tests/test_bootstrap.py
+++ b/tests/test_bootstrap.py
@@ -109,7 +109,7 @@ async def test_bootstrap_license(
                         "secretKeyRef": {"key": "license", "name": f"license-{name}"},
                     },
                     "name": "my-crate-cluster",
-                    "version": "4.1.5",
+                    "version": "4.3.3",
                 },
                 "nodes": {
                     "data": [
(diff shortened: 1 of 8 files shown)


lukasbals (approved):
Looks good to me 👍🏽

tests/test_bootstrap.py
```diff
@@ -109,7 +109,7 @@ async def test_bootstrap_license(
                         "secretKeyRef": {"key": "license", "name": f"license-{name}"},
                     },
                     "name": "my-crate-cluster",
-                    "version": "4.1.5",
+                    "version": "4.3.3",
```
turbo-ele, line 112: note: `4.3.4` would be the latest `4.3.x` release, while `4.4.0` is the latest stable

SStorm: Good point, I will change it to 4.4.0 in a subsequent PR if that's ok.

lukasbals: Haha true. I just approved a few minutes before `4.4.0` was out. 😂

PR #176, opened 2021-03-11
Missing await meant we're not sending any notifications

SStorm:
## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,11 @@
 Changelog
 =========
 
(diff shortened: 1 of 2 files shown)


lukasbals:
Good catch! But I think we should add a changelog entry.

lukasbals (approved):

SStorm:
> Good catch! But I think we should add a changelog entry.

Whoops thanks. The missing checklist in the PR template keeps tripping me up.

Issue #61, opened 2020-08-26
Cluster restart due to update lead to broken snapshots


turbo-ele:
One of our clusters got updated from `4.2.2` to `4.2.3` and (possibly) restarted in the middle of a snapshot. This lead to all following (incremental) snapshots being broken:

![image](https://user-images.githubusercontent.com/5152633/91268759-2a4bf680-e776-11ea-909f-945d8887f8a6.png)

Excerpt from failures column:
```
 Array[210]
1
:
[...]

PR #184, opened 2021-03-24
Prevent scaling/upgrading/restarting when a snapshot/backup is in progress

SStorm:

Before performing scaling/upgrading/restarts, check that:
1. There are any `CREATE SNAPSHOT ...` jobs in CrateDB. If so, waits for them to finish.
2. There is a running k8s job to perform the backup. If so, waits for it to finish. <- Note that this has to be done in addition to the `CREATE SNAPSHOT` check because there is a real possibility of a race condition - it might take up to 30s for the `CREATE SNAPSHOT` to appear once the backup job starts.

This is implemented as an additional state in the cluster update state machine - basically we block in the `ensure_no_backups` state until there are no backups running any more.

Also refactored the ITs somewhat to better reuse code (especially cluster creation).

Also fixed a bug that meant we're not waiting for a cluster to become GREEN before performing scaling (caused by a missing `await`).

## Checklist

### DIFF
--- a/crate/operator/cratedb.py
+++ b/crate/operator/cratedb.py
@@ -171,6 +171,17 @@ async def is_cluster_healthy(
 async def are_snapshots_in_progress(
     conn_factory, logger: logging.Logger
 ) -> Tuple[bool, Optional[str]]:
+    """
+    Check if there are any snapshots in progress by querying sys.jobs
+    (or the configured table for testing purposes).
+
+    Since sys.jobs will also contain the current query (i.e. us selecting from sys.jobs)
+    we also need to exclude our statement.
+
+    :param conn_factory the connection factory to connect to CrateDB
+    :param logger the logger on which we're logging
+    :return A Tuple of the result (Bool) and the statement that is progress (if any).
+    """
     async with conn_factory() as conn:
         async with conn.cursor() as cursor:
             logger.info("Checking if there are running snapshots ...")
--- a/crate/operator/cratedb.py
+++ b/crate/operator/cratedb.py
@@ -16,12 +16,14 @@
 
 import functools
 import logging
+from typing import Optional, Tuple
 
 import aiopg
 from aiopg import Cursor
 from psycopg2 import ProgrammingError
 from psycopg2.extensions import quote_ident
 
+from crate.operator.config import config
 from crate.operator.constants import SYSTEM_USERNAME
 
 HEALTHINESS = {1: "GREEN", 2: "YELLOW", 3: "RED"}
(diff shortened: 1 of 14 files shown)


lukasbals (approved):
Good work!

turbo-ele:
Related issue: https://github.com/crate/crate-operator/issues/61

crate/operator/cratedb.py
```diff
@@ -164,3 +166,22 @@ async def is_cluster_healthy(
...
+
+async def are_snapshots_in_progress(
+    conn_factory, logger: logging.Logger
+) -> Tuple[bool, Optional[str]]:
+    async with conn_factory() as conn:
+        async with conn.cursor() as cursor:
+            logger.info("Checking if there are running snapshots ...")
+            try:
+                await cursor.execute(
+                    f"SELECT stmt FROM {config.JOBS_TABLE} WHERE stmt "
```
turbo-ele, line 179: is there a reason, why you don't use the state of the sys.snapshots table?
https://crate.io/docs/crate/reference/en/4.4/admin/system-information.html#snapshots

crate/operator/cratedb.py
```diff
@@ -164,3 +166,22 @@ async def is_cluster_healthy(
             except ProgrammingError as e:
                 logger.warning("Failed to run health check query", exc_info=e)
                 return False
+
+
+async def are_snapshots_in_progress(
+    conn_factory, logger: logging.Logger
+) -> Tuple[bool, Optional[str]]:
```
turbo-ele, line 173: Just a note: personally, i would document what the return values are about. especially the second is not obvious.

turbo-ele (approved):
lgtm 😎

crate/operator/cratedb.py
SStorm: Yes, a very good reason - I didn't know/remember it! :) But this is much better, I will change.

SStorm: Update: `sys.snapshots` does not show _in progress_ snapshots - only finished ones. So I cannot use this.

mfussenegger: Since CrateDB 4.4.0 there would be a `snapshot_restore` table, see https://github.com/crate/crate/commit/473400e096eb08c2cf12d590075627b6fe904157

SStorm: Oh, thanks @mfussenegger! This is great. We might want to change it to use that in the future, once we have no clusters < 4.4.0. If no one minds, for now I'd like to keep the query from `sys.jobs`, as it works well enough.

PR #188, opened 2021-03-30
Disable the backup CronJob when performing cluster updates

SStorm:

This does quite a few things. The gist of it is to disable & *re-enable* the backup cronjob if we are performing update operations on the cluster. There is no easy way to do this using kopf handlers, because there's no concept of a finalising handler that runs no-matter what. 

We want to do this because running a backup while an update is in progress and nodes are added/removed/restarted can potentially make the backup stuck, which requires manual intervention from a sysadmin. 

To implement this, I have done several things:
1. Completely removed the state machine that was there before, instead opting for a dependency mechanism between kopf sub-handlers. The SM seemed over-complicated for this. For each running sub-handler you can specify what it's dependent handlers are, and it'll keep failing until the dependencies are met.
2. Added an ability for some handlers to run even if their dependent handler has failed - this is then used to ensure that the cronjob is re-enabled no matter what. Even if the cluster upgrade/scale has failed. 
3. Added a hash of the kopf diff (i.e. what are we about to change in the cluster) to the status, so that we can keep track of which run is currently happening. The mechanism for that is better described  [here](https://github.com/crate/crate-operator/pull/188/files#diff-bb3b46ad401b4c8862115080ed962acece1010805eb1316e0b0e7fdf5c4c8952R397).

## Checklist


CHANGES.rst
```diff
@@ -20,6 +20,16 @@ Unreleased
 * Changed how (sub)handlers are treated to allow returning statuses, which get persisted
   against the CrateDB resource in k8s.
 
+* Changed cluster updates to disable any backup cronjobs, so that a job doesn't
+  kick in just as we are performing a cluster update. The job will be re-enabled
+  once the update is complete
```
matthijskrul, line 25: . missing :)

crate/operator/utils/kopf.py
```diff
@@ -42,45 +43,75 @@ async def _wrapper(**_):
...
+        :param depends_on: list of dependent handler this handler should wait for.
+        :param run_on_dep_failures: whether we should still execute if our dependencies
+            have failed. This is useful for handlers that clean up resources after
+            other handlers have finished, and always need to run.
+        """
         self.namespace = namespace
         self.name = name
+        self.ref = ref
         self._context = context
+        self.depends_on = depends_on if depends_on is not None else []
```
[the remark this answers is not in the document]
SStorm: Good point, but I did it this way since PyCharm complained:
Screenshot from 2021-04-08 16-09-08

:shrug: ?

crate/operator/backup.py
```diff
@@ -326,26 +383,95 @@ async def handle(  # type: ignore
...
+                    )
                     raise kopf.TemporaryError(
                         "A snapshot k8s job is currently running, "
                         f"waiting for it to finish: {job_name}",
                         delay=30,
                     )
+
+    async def _notify_backup_running(self, logger):
+        ...
+        # TODO: Send notification. This throws an exception as the API validation fails.
```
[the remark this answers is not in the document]
SStorm: Yes, it's still the trello card's TODO list.

crate/operator/utils/kopf.py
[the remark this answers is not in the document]
Taliik: Okay, this might have been a bad suggestions, let's not use mutable values as default values, especially not lists as this might cause issues later on 😅

lukasbals (approved):
LGTM 👍🏽 Good work

crate/operator/main.py
```diff
@@ -380,20 +385,54 @@ async def cluster_create(
             )
 
 
-@kopf.on.update(API_GROUP, "v1", RESOURCE_CRATEDB)
+@kopf.on.update(API_GROUP, "v1", RESOURCE_CRATEDB, id="cluster_update")
```
turbo-ele, line 388: optional: you could also use a const for the id `cluster_update`

crate/operator/main.py
```diff
@@ -380,20 +385,54 @@ async def cluster_create(
...
+    the hash matches - if not, it means we can disregard any refs that are not for this
+    run.
+    """
+    context = status.get("cluster_update")
+    hash = hashlib.md5(str(diff).encode("utf-8")).hexdigest()
+    if not context:
+        context = {"ref": hash}
+
+    if context.get("ref", "") != hash:
+        context["ref"] = hash
```
turbo-ele, line 435:
```suggestion
    if not context:
        context = {"ref": hash}
    elif context.get("ref", "") != hash:
        context["ref"] = hash
```

crate/operator/utils/kopf.py
```diff
@@ -42,45 +43,75 @@ async def _wrapper(**_):
...
+        name: str,
+        ref: str,
+        context: dict,
+        depends_on=None,
+        run_on_dep_failures=False,
+    ):
+        """
+        Constructs a new dependency-aware handler.
+
+        :param namespace: the namespace to use
```
turbo-ele, line 77: `ref` missing in docstring param description

crate/operator/utils/kopf.py
```diff
@@ -107,7 +140,51 @@ async def send_notifications(self, logger: logging.Logger):
...
+
+        What we are doing instead is peeking into kopf's internal state storage -
+        the annotations on the CrateDB objects to check if the handler has failed.
+
+        Slightly naughty, but there is no better way at the time of writing.
+        """
+        # Handler names have dots instead of slashes in annotations
+        normalized_name = handler_name.replace("/", ".")
+        key = f"{KOPF_STATE_STORE_PREFIX}/{normalized_name}"
+        status_str = annotations.get(key, None)
```
turbo-ele, line 175: `None` is the default of get
```suggestion
        status_str = annotations.get(key)
```

turbo-ele (approved):
looks nice 👍

PR #189, opened 2021-04-01
Allow returning a status from handlers

SStorm:

Handlers have a facility to store their status. That is stored in the "status" attribute of the resource that is being handled (CrateDB in our case). This was not working before, but is quite useful.

As a result of this change, some handlers had to be adapter to not return large representations (i.e. entire K8s resource objects), because these need to be serialiased and stored against the CrateDB resource (which obviously doesn't work).

The usage of these statuses allows some control of handler execution sequencing, because it is possible to check if a previous handler has succeeded or not, as described [here](https://kopf.readthedocs.io/en/latest/results/).

## Checklist

### DIFF
--- a/crate/operator/create.py
+++ b/crate/operator/create.py
@@ -18,7 +18,7 @@ import logging
 import math
 import pkgutil
 import warnings
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional
 
 import bitmath
 import yaml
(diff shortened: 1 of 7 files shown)


crate/operator/create.py
```diff
@@ -128,7 +128,7 @@ async def create_debug_volume(
     name: str,
     labels: LabelType,
     logger: logging.Logger,
-) -> Tuple[V1PersistentVolume, V1PersistentVolumeClaim]:
+) -> None:
```
SStorm, line 131: Note: everywhere where the return type has been changed to `None`, the returned values have not been used for anything.

crate/operator/utils/kopf.py
```diff
@@ -37,7 +37,7 @@ def subhandler_partial(awaitable: Callable, *args, **kwargs):
 
     @wraps(awaitable)
     async def _wrapper(**_):
-        await awaitable(*args, **kwargs)
+        return await awaitable(*args, **kwargs)
```
SStorm, line 40: This is the actual change that allows having statuses.

Taliik (approved):
Lgtm 🚀

lukasbals (approved):

PR #192, opened 2021-04-12
Add PDB for statefulset

WalBeh:
## Why this is an improvement
A Pod Distruption Budget ensures that when performing Kubernetes cluster maintenance (i.e. upgrades), we make sure to not disrupt more than 1 pod in a StatefulSet at a time. 

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,7 @@ Unreleased
 
 * Removed the Context class in favour of simple storing the context as a dictionary.
 
+* Added PodDisruptionBudget to keep a cratedb statefulset up during k8s upgrades.
 
 1.2.0 (2021-03-22)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changelog
 Unreleased
 ----------
 
+* Added PodDisruptionBudget to keep a cratedb statefulset up during kubernetes upgrades.
+
 * Added a check for any running snapshots (either k8s jobs or CREATE SNAPSHOT stmts.)
   before performing scaling/upgrades/restarts. This ensures we don't inadvertently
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -29,6 +29,7 @@ Unreleased
 
 * Removed the Context class in favour of simple storing the context as a dictionary.
 
+* Added PodDisruptionBudget to keep a cratedb statefulset up during k8s upgrades.
```
turbo-ele, line 32: We usually add new entries on top of the list

```suggestion
* Added PodDisruptionBudget to keep a CrateDB StatefulSet up during Kubernetes upgrades.
```

crate/operator/create.py
```diff
@@ -768,6 +771,30 @@ async def create_statefulset(
                 logger,
             ),
         )
+        policy = PolicyV1beta1Api(api_client)
```
SStorm, line 774: I think lets add a comment here saying what a PDB is, as it's not super clear what it is for w/o reading the docs.

Something along the lines of "A Pod Distruption Budget ensures that when performing k8s cluster maintenance (i.e. upgrades), we make sure to not disrupt more than X pods in a StatefulSet at a time. This means that when doing a rolling restart, we'll wait for all pods to come back before restarting the next k8s node."

lukasbals (approved):
Agree with the comment from Romanas. Otherwise LGTM 👍🏽

turbo-ele (approved):
LGTM! 👍 
One general question regarding compatibility. As far as I've seen PDB is supported in Kubernetes versions >1.5 as beta feature. Since 1.21 it also has been declared stable, should we address this somehow?

SStorm (approved):

juanpardo (approved):

PR #197, opened 2021-04-22
Enable sending delay notifications

Taliik:
Send delay notifications when a snapshot is in progress and a operation
(scale/upgrade/restart) is triggered

## Checklist

### DIFF
--- a/crate/operator/backup.py
+++ b/crate/operator/backup.py
@@ -57,6 +57,11 @@ from crate.operator.utils.kubeapi import (
     get_system_user_password,
 )
 from crate.operator.utils.typing import LabelType
+from crate.operator.webhooks import (
+    WebhookEvent,
+    WebhookStatus,
+    WebhookTemporaryFailurePayload,
+)
 
 
 def get_backup_env(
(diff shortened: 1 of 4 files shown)


crate/operator/utils/kopf.py
```diff
@@ -119,10 +119,11 @@ async def handle(self, **kwargs: Any):
         raise NotImplementedError()
 
     def schedule_notification(
-        self, event: WebhookEvent, payload: WebhookSubPayload, status: WebhookStatus
+        self, event: WebhookEvent, payload: WebhookSubPayload, status: WebhookStatus,
+
     ):
         self._context.setdefault("notifications", []).append(
-            Notification(event=event, payload=payload, status=status)
+            Notification(event=event, payload=payload, status=status, )
```
SStorm, line 126: comma not required here?..

crate/operator/webhooks.py
```diff
@@ -275,5 +281,31 @@ async def send_upgrade_notification(
             logger=logger,
         )
 
+    async def send_delay_notification(
```
SStorm, line 284: I just realised that `send_delay_notification`, `send_scale_notification` and `send_upgrade_notification` are only used in tests. I wonder if we could use the generic `send_notification` method everywhere and get rid of these altogether?..

Taliik: Yeah I wondered the same, but since for the other operations we had these specific functions I added the same for `delay`. I tested it with the `send_notification` and it worked too, might be worth to remove these "testing" only functions and just use `send_notification` for all.

SStorm: That might be a good idea, up to you if you want to do it as part of this PR though!

crate/operator/backup.py
```diff
@@ -390,13 +395,6 @@ async def _ensure_no_snapshots_in_progress(self, namespace, name, logger):
                 # Raising a TemporaryError will clear any registered subhandlers, so we
```
SStorm, line 395: This comment is no longer required now.

SStorm (approved):

PR #214, opened 2021-06-14
Send a webhook when an external IP is obtained

SStorm:

This is required in the cloud to update DNS for Edge regions which do not run external-dns.

## Checklist

### DIFF
--- a/crate/operator/main.py
+++ b/crate/operator/main.py
@@ -21,6 +21,7 @@ import logging
 from typing import Any, Awaitable, Dict, List
 
 import kopf
+from kopf.structs.diffs import DiffItem, DiffOperation
 from kubernetes_asyncio.client import (
     CoreV1Api,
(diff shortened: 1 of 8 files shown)


crate/operator/main.py
```diff
@@ -550,3 +552,48 @@ async def secret_update(
                             id=f"update-{crate_custom_object['metadata']['name']}-{user_spec['name']}",  # noqa
                             timeout=config.BOOTSTRAP_TIMEOUT,
                         )
+
+
+@kopf.on.field(
+    "",
+    "v1",
+    "services",
+    labels={LABEL_PART_OF: "cratedb", LABEL_MANAGED_BY: "crate-operator"},
```
SStorm, line 561: Filtering for resources that are managed by the operator ONLY so we don't pick up random other services.

joncombe (approved):

PR #235, opened 2021-08-03
Add additional env variable to cronjobs for custom s3 backup endpointUrl

WalBeh:
This adds additional environment variables to the backup cronjob to be able to write to a custom S3 bucket.

## Checklist

### DIFF
--- a/deploy/crd.yaml
+++ b/deploy/crd.yaml
@@ -78,6 +78,23 @@ spec:
                       description: A crontab formatted string indicating when and how often to perform backups.
                       pattern: ^(((\d+,)+\d+|(\d+(\/|-)\d+)|\d+|\*) ){4}(((\d+,)+\d+|(\d+(\/|-)\d+)|\d+|\*))$
                       type: string
+                    endpointUrl:
+                      properties:
+                        secretKeyRef:
+                          properties:
+                            key:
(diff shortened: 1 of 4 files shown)


deploy/crd.yaml
```diff
@@ -109,6 +109,34 @@ spec:
                           - key
                           - name
                           type: object
+                    endpoint:
+                      properties:
```
SStorm, line 113: I think this is trying to say that it has to be alphabetical...

deploy/crd.yaml
```diff
@@ -78,6 +78,37 @@ spec:
...
+                            key:
+                              description: The key within the Kubernetes Secret that holds the S3 endpoint.
+                              type: string
+                            name:
+                              description: Name of a Kubernetes Secret that contains the S3 endpoind Access Key to be used for backups.
+                              type: string
+                          required:
+                          - key
+                          - name
+                          type: object
```
SStorm, line 94: I think you also want this part here too:
```
required:
- secretKeyRef
```

SStorm (approved):
Looks good! I think there's a minor missing part that I commented, but it doesn't change much.

SStorm (approved):
:100:

PR #251, opened 2021-10-04
Use new gateway settings from crate version 4.7 onwards

tomach:

CrateDB settings gateway.expected_nodes and gateway.recover_after_nodes will be deprecated in version 5, from 4.7 onwards the new settings gateway.expected_data_nodes and gateway.recover_after_data_nodes can be used instead. This change affects the following scenarios:

- create: depending on the version, the required settings' names and respective values are used

- upgrade: usually the settings' names and number of nodes do not change. Only when there is an upgrade from version < 4.7 to >= 4.7 the command of the sts needs to be retrieved, the settings names must be replaced and the number of data nodes must be used instead of the number of total nodes.

- scale: after 4.7 only the number of data nodes are considered when updating the sts commands

## Checklist

 - [X] [CLA](https://crate.io/community/contribute/cla/) is signed


crate/operator/operations.py
```diff
@@ -57,7 +57,7 @@
 )
 
 
-def get_total_nodes_count(nodes: Dict[str, Any]) -> int:
+def get_total_nodes_count(nodes: Dict[str, Any], type: str = "all") -> int:
```
SStorm, line 60: Nice refactor! Thanks.

SStorm (approved):
Nice work! I suppose technically we don't have to use these changes until 5.0, but it's nice to have a mechanism for them before that. A bit of a shame we have to have special handling, but what can we do :/

SStorm (approved):

PR #254, opened 2021-10-06
Added service IP Whitelist to create.py

WalBeh:
This allows to add IP addresses in CIDR notation to the crate service. Once defined, it allows connection to cratedb only from the specified IP ranges. 

cloud-234

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -753,6 +753,41 @@ class TestServices:
             namespace.metadata.name,
             {f"crate-{name}", f"crate-discovery-{name}"},
         )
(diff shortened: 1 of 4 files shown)


tests/test_create.py
```diff
@@ -754,6 +754,37 @@ async def test_create(self, faker, namespace, api_client):
             {f"crate-{name}", f"crate-discovery-{name}"},
         )
```
SStorm, line 755: I think it would be good to add this here:
```
        assert service.spec.load_balancer_source_ranges is None
```

PR #278, opened 2021-11-01
Expose a Prometheus metrics endpoint

SStorm:

By default, on port 8080.

This also exposes cluster health metrics for each running cluster.

Once deployed, we can use this for the `CloudClusterHealthStatusError` and `CloudClusterLastSeenTooOld` alerts instead of using the API metrics.

## Checklist


tomach (approved):
Awesome :+1:

PR #289, opened 2021-11-10
Rework cluster health metric gathering

SStorm:
## Summary of changes

Prior to this change, metrics would "hang around" even if a cluster is deleted. This created a problem as deleted clusters would then trigger the LastSeenTooOld alert and 24x7.

The new method instead has a metric gatherer, which only reports metrics for clusters that exist. If a cluster is deleted, it's metrics will disappear in a few minutes and we won't get alerted.

## Checklist

### DIFF
--- a/crate/operator/prometheus.py
+++ b/crate/operator/prometheus.py
@@ -1,6 +1,26 @@
-from datetime import datetime
+# CrateDB Kubernetes Operator
+# Copyright (C) 2021 Crate.IO GmbH
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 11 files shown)


crate/operator/prometheus.py
```diff
@@ -13,14 +33,51 @@
     }
 )
 
-cluster_status_gauge = Gauge(
-    "cloud_clusters_health",
-    documentation="0->GREEN, 1->YELLOW, 2->RED, 3->UNREACHABLE",
-    labelnames=["cluster_id"],
-)
+CLUSTER_METRICS = {}
```
SStorm, line 36: tldr; version of the change - we keep the metrics in-memory in this dictionary, which is updated by the `report_cluster_status` function, which is in turn called by the `handle_ping_cratedb_status.py ` handler. If a cluster is deleted, the handler will no longer run and the metric for this cluster will timeout (`if now - metrics["last_reported"] < LAST_SEEN_THRESHOLD:`) and no longer be returned.

If anyone has a better idea, I'd love to hear it!

tomach: sounds good. I don't think I have a better idea :)

tomach (approved):

PR #291, opened 2021-11-11
Add notifications for exceptions

tomach:
## Summary of changes
This is a proposal how we could handle notifications to the API in case of exceptions or timeouts during an upgrade (or all other operations in the future). 

Added two `wrapt` universal decorators which can be applied to plain functions (for regular kopf handlers) or class methods (for subhandlers using StateBasedSubHandler). `@crate.timeout(timeout={seconds})` checks if the (sub)handler's runtime exceeded the given timeout seconds on each run and raises a `kopf.HandlerTimeoutError` in case it has timed out. This exception as well as all other `kopf.PermanentError`s are caught in `@crate.on.error(error_handler={my_error_handler})` and an error handler coroutine can be run sending notifications to the API. After that it re-raises the exception so it can be further processed by kopf.

This PR also fixes the issue that notifications of successful upgrades pile up in the status of the CrateDB resource if an upgrade succeeds but the subsequent restart fails or times out. These notifications were erroneously sent in the next run of the handler, because on permanent errors or timeouts there is no traditional return value and kopf does not update the status in such cases. After two failed restarts it looks sth like this

```yml
[...]
status:
  cluster_update:
    notifications:
    - event: upgrade
      payload:
        new_registry: crate
        new_version: 4.6.3
        old_registry: crate
        old_version: 4.6.1
      status: success
    - event: upgrade
      payload:
        new_registry: crate
        new_version: 4.6.4
        old_registry: crate
        old_version: 4.6.3
      status: success
    ref: 20681de22617dfdd8ef03effd25b8014
  cluster_update/after_cluster_update:
    ref: 20681de22617dfdd8ef03effd25b8014
    success: true
[...]
```
I added a new subhandler which only creates the success notification if both `upgrade` and `restart` subhandlers have finished successfully, not sure if there would be a more elegant way of solving this issue..

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,21 @@ Changelog
 Unreleased
 ----------
 
+* Replaced kopf timeout handling with a decorator ``@crate.timeout()`` to be
+  able to run code when a timeout happens.
+
+* Added a decorator ``@crate.on.error()`` which catches timeouts as well as
+  other permanent handler errors and performs actions passed in an error
+  handler, like sending a notification.
+
+* Fixed the issue that notifications of successful upgrades pile up in the
+  status of the CrateDB resource if an upgrade succeeds but the subsequent
+  restart fails or times out. These notifications were erroneously sent in the
+  next run of the handler.
+
+* Changed the registration of all kopf subhandlers in the creation process
+  to use StateBasedSubhandler.
+
 2.7.1 (2021-11-12)
 ------------------
 
--- a/crate/operator/main.py
+++ b/crate/operator/main.py
@@ -48,6 +48,7 @@ from crate.operator.handlers.handle_update_user_password_secret import (
     update_user_password_secret,
 )
 from crate.operator.kube_auth import login_via_kubernetes_asyncio
+from crate.operator.utils import crate
 from crate.operator.webhooks import webhook_client
 
 NO_VALUE = object()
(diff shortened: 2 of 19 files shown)


CHANGES.rst
```diff
@@ -5,6 +5,18 @@ Changelog
 Unreleased
 ----------
 
+* Replaced kopf timeout handling with a decorator ``@crate.timeout()`` to be
+  able to handle timeouts in the operator.
```
SStorm, line 9:
```suggestion
* Replaced kopf timeout handling with a decorator ``@crate.timeout()`` to be
  able to run code when a timeout happens.
```

crate/operator/main.py
```diff
@@ -105,6 +106,7 @@ async def cluster_create(
 
 
 @kopf.on.update(API_GROUP, "v1", RESOURCE_CRATEDB, id=CLUSTER_UPDATE_ID)
+@crate.on.error(error_handler=crate.send_error_notification)
```
SStorm, line 109: Would it be nicer to send a error notification for the specific operation that failed? i.e. instead of `WebhookEvent.ERROR` with `WebhookStatus.FAILURE` send a `WebHookEvent.UPGRADE` with `WebhookStatus.FAILURE`? 

Just thinking that it would be easier to map this to a running operations in the API then?..

```suggestion
@crate.on.error(error_handler=crate.send_upgrade_failed_notification)
```

crate/operator/utils/kopf.py
```diff
@@ -83,17 +86,21 @@ def __init__(
...
         self.ref = ref
         self._context = context
         self.depends_on = depends_on if depends_on is not None else []
         self.run_on_dep_failures = run_on_dep_failures
+        self.timeout = timeout
 
     def __call__(self, **kwargs: Any):
         return functools.partial(self._subhandler, **kwargs)
 
+    @crate.on.error(error_handler=crate.send_error_notification)
```
SStorm, line 102: Should we put these here or on specific `handle()` methods in each handler? Then you can fine-tune it a bit more and send specific failure notifications for specific things that failed?...

crate/operator/upgrade.py
```diff
@@ -173,10 +174,38 @@ async def handle(  # type: ignore
         logger: logging.Logger,
         **kwargs: Any,
     ):
+        self.schedule_notification(
```
SStorm, line 177: Is it worth moving this to _after_ `await upgrade_cluster`?

SStorm (approved):
:100:

crate/operator/upgrade.py
tomach: sure, we can do that.

PR #303, opened 2021-12-10
Add timeouts to update and create handlers

tomach:
## Summary of changes
https://github.com/crate/cloud/issues/429 - Add timeouts to the main `update` and `create` handlers to make sure operations are not blocked forever by a subhandler which does not have an individual timeout.

## Checklist

### DIFF
--- a/crate/operator/main.py
+++ b/crate/operator/main.py
@@ -97,6 +97,7 @@ async def login(**kwargs):
 
 @kopf.on.create(API_GROUP, "v1", RESOURCE_CRATEDB)
 @crate.on.error(error_handler=crate.send_create_failed_notification)
+@crate.timeout(timeout=float(config.BOOTSTRAP_TIMEOUT))
 async def cluster_create(
     namespace: str,
     meta: kopf.Meta,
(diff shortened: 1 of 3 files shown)


crate/operator/main.py
```diff
@@ -114,6 +115,7 @@ async def cluster_create(
 
 @kopf.on.update(API_GROUP, "v1", RESOURCE_CRATEDB, id=CLUSTER_UPDATE_ID)
 @crate.on.error(error_handler=crate.send_update_failed_notification)
+@crate.timeout(timeout=float(config.ROLLING_RESTART_TIMEOUT))
```
SStorm, line 118: I am a bit uneasy about setting this to `ROLLING_RESTART_TIMEOUT` as it is set to 1 hour - which might not be enough in cases where we are, i.e., waiting for a backup to finish. But we can probably go with this and see how much of a problem it is

SStorm (approved):

crate/operator/main.py
tomach: We decided to add a new config option for the entire update process and set it to 2 hours for now.

SStorm (approved):

PR #327, opened 2022-02-21
Headless discovery service

SStorm:
## Summary of changes

Somehow accidentally we've had the `crate-discovery` service load balanced. There is no reason for it. In fact, cluster discovery worked only by chance since the nodes would _eventually_ be load balanced to a node they haven't spoken with yet - which is silly.

Have also removed a bunch of `@kopf.on-resume` handlers that make no sense any more - everything has long been migrated.

Ref https://github.com/crate/cloud/issues/513

## Checklist


crate/operator/create.py
```diff
@@ -845,6 +845,8 @@ def get_discovery_service(
             owner_references=owner_references,
         ),
         spec=V1ServiceSpec(
+            # Headless service
+            cluster_ip="None",
```
SStorm, line 849: A quick explanation of what is happening here:

Setting the `cluster_ip` explcitly to "None" (the String `"None"`, not the null `None`) will mean that kubernetes will not create an internal load balancer for it. Instead, when any pods try to access this service, the DNS record for it will return multiple IP addresses - the IPs of the individual pods that this service is "fronting". 

As far as crate is concerned, this is great for service discovery - it means all the crate nodes immediately see all the other nodes, instead of randomly connecting via a load balancer and by chance ending up on one they don't know (yet).

WalBeh (approved):

PR #346, opened 2022-04-01
Add requests/limits for cluster resources to crd

tomach:
## Summary of changes
Basically the idea is, to make `resources.cpus` and `resources.memory` not mandatory anymore in the crd for cratedb nodes. Instead mark them as deprecated and add `resources.requets.(cpu|memory)` and `resources.limits.(cpu|memory)` to the crd. When creating the STS, check if both `resources.requets.(cpu|memory)` and `resources.limits.(cpu|memory)` are set AND different -> assume that it is a cluster on shared resources then, and set `nodeAffinity` instead of `podAntiAffinity` and `tolerations` according to the Proposed implementation from https://github.com/crate/cloud/issues/522.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: toleration-nodelabel
  namespace: default
spec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: cratedb
            operator: In
            values:
            - shared
  containers:
  - image: k8s.gcr.io/pause:2.0
    name: toleration-check-affinity
    resources:
      requests:
        cpu: 100m
        memory: 128Mi
      limits:
        cpu: 1
        memory: 2G
  tolerations:
  - key: "cratedb"
    operator: "Equal"
    value: "shared"
    effect: "NoSchedule"
```

## Checklist


WalBeh (approved):
LGTM

crate/operator/create.py
```diff
@@ -152,30 +157,68 @@ async def create_sql_exporter_config(
...
-                        ],
+    if is_shared_resources_cluster(node_spec):
+        return V1Affinity(
+            node_affinity=V1NodeAffinity(
+                required_during_scheduling_ignored_during_execution=V1NodeSelector(
+                    node_selector_terms=[
+                        V1NodeSelectorTerm(
+                            match_expressions=[
+                                V1NodeSelectorRequirement(
+                                    key="cratedb", operator="In", values=["shared"]
```
SStorm, line 175: I think it would be good if these were constants somewhere, just like we have it in the `V1Affinity` section below.

crate/operator/create.py
```diff
@@ -963,6 +1031,56 @@ async def create_system_user(
...
+            (
+                node_spec["resources"].get("limits", {}).get("memory")
+                and node_spec["resources"].get("requests", {}).get("memory")
+                and node_spec["resources"].get("limits", {}).get("memory")
+                != node_spec["resources"].get("requests", {}).get("memory")
+            )
+            or False
+        )
+    except KeyError:
+        return False
```
SStorm, line 1054: Could this be simplified? Considering that for it to be a shared cluster, the requests must be set and they must be different from limits, we could perhaps do something like this:
```suggestion
    cpu_req = node_spec["resources"].get("requests", {}).get("cpu")
    cpu_limit = node_spec["resources"].get("limits", {}).get("cpu")
    mem_req = node_spec["resources"].get("requests", {}).get("memory")
    mem_limit = node_spec["resources"].get("limits", {}).get("memory")
    if not cpu_req or mem_req:
        return False
    return cpu_req != cpu_limit or mem_req != mem_limit
```

Maybe a bit more readable? But up to you!

crate/operator/create.py
```diff
@@ -408,7 +473,9 @@ def get_statefulset_crate_env(
             name="CRATE_HEAP_SIZE",
             value=str(
                 int(
-                    bitmath.parse_string_unsafe(node_spec["resources"]["memory"]).bytes
+                    bitmath.parse_string_unsafe(
+                        get_cluster_resource_limits(node_spec, resource_type="memory")
```
SStorm, line 477: Should this have a fallback key too?

SStorm (approved):
Nice work! A few suggestions for you to consider.

crate/operator/create.py
tomach: No, for `memory` the keys are the same, it was only required for `cpu` vs `cpus` (the old `resources.cpus` and now `resources.limits.cpu`)

PR #358, opened 2022-05-31
Suspend clusters support

juanpardo:
## Checklist

### DIFF
--- a/crate/operator/scale.py
+++ b/crate/operator/scale.py
@@ -567,14 +567,12 @@ async def scale_cluster(
                     ):
                         new_num_data_nodes = total_number_of_nodes - num_master_nodes
 
-                    # Only deallocate nodes if the cluster is not being suspended
-                    if new_num_data_nodes > 0:
-                        await deallocate_nodes(
-                            conn_factory,
-                            new_num_data_nodes,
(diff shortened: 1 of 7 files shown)


crate/operator/scale.py
```diff
@@ -663,16 +663,42 @@ async def handle(  # type: ignore
...
+            if (
+                # Ensure the diff data exists to keep mypy happy
+                scale_data_diff_items
+                and len(scale_data_diff_items[0]) >= 4
+                # Check for suspend or resume
+                and (
+                    scale_data_diff_items[0][2] == 0 or scale_data_diff_items[0][3] == 0
+                )
+            ):
+                await suspend_or_start_cluster(
```
tomach, line 677: This means that suspending a cluster sends the `IN_PROGRESS` status notification with the text `Suspending cluster and waiting for Persistent Volume Claim(s) to be resized.` to the API [here](https://github.com/crate/crate-operator/blob/master/crate/operator/operations.py#L526), right? Probably we should update this text.

For scaling up it is `Starting cluster. Scaling back up to {new_replicas} nodes. Waiting for node(s) to be present.` which should be fine I think.

PR #364, opened 2022-06-10
Suspending a cluster now sends the right message when in progress

juanpardo:
## Summary of changes
Suspending a cluster now sends the right message when in progress

## Checklist


crate/operator/expand_volume.py
```diff
@@ -184,3 +187,104 @@ async def handle(  # type: ignore
             ),
             WebhookStatus.SUCCESS,
         )
+
+
+class StartClusterSubHandler(StateBasedSubHandler):
```
tomach, line 192: I think we moved `StartClusterSubHandler` and `SuspendClusterSubHandler` intentionally to the generic file back then to be able to reuse it from different operations.

https://github.com/crate/crate-operator/pull/330#pullrequestreview-914265413

> Do you think it might be useful to re-structure things a bit and move these generic handlers into a separate file? As they have nothing to do with expanding volumes per se...

juanpardo: I had to move them due to a circular dependency, but will try to overcome it and keep them where they were!

plaharanne (approved):
👍

PR #371, opened 2022-06-24
Fix a bug that would lead to the operator getting stuck when performing repeated operations

SStorm:
## Summary of changes

This was caused by the hash staying the same and the operator failing to figure out what ops we still need to wait for.

## Checklist


Taliik (approved):

tomach (approved):
:+1:

PR #379, opened 2022-07-11
Fix a bug that would cause suspending a cluster to get stuck

tomach:
## Summary of changes
Fixed a bug in the check if the backup metrics deployment has been scaled down successfully, or rather if the `backup-metrics` pod is gone. The `crate-snapshot` pods in the same namespace have the same labels, so the operation to suspend a cluster would have been stuck as long as such a pod was present

```
Handler 'cluster_update/scale' failed temporarily: Waiting for backup metrics pods [{'uid': '719c8f9f-743c-497e-aceb-7b46ba98edeb', 'name': 'create-snapshot-7987a175-ea95-445d-995f-43378fe3a16d-276252qd6z'}] to match desired number of replicas 0
```

https://github.com/crate/cloud/issues/702

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changelog
 Unreleased
 ----------
 
+* Fix a bug that would cause suspending a cluster to get stuck.
+
 2.13.2 (2022-07-11)
 -------------------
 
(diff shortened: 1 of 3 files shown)


SStorm (approved):
Nice catch

PR #386, opened 2022-08-10
Added support to change cpu and memory on running clusters

juanpardo:
## Checklist


crate/operator/change_plan.py
```diff
@@ -0,0 +1,121 @@
+import logging
+from typing import Any
+
+import kopf
+from kubernetes_asyncio.client import AppsV1Api
+from kubernetes_asyncio.client.api_client import ApiClient
+from webhooks import WebhookChangePlanPayload, WebhookEvent, WebhookStatus
```
SStorm, line 7:
```suggestion
from crate.operator.webhooks import WebhookChangePlanPayload, WebhookEvent, WebhookStatus
```

This is what's causing the docs build to fail.

crate/operator/change_plan.py
```diff
@@ -0,0 +1,121 @@
+import logging
```
SStorm, line 1: I think we should avoid calling it "plan" anywhere as the operator has no such concept - as far as it's concerned, we're just changing the compute. How about `change_compute`, `CHANGE_COMPUTE` etc?

crate/operator/change_plan.py
```diff
@@ -0,0 +1,121 @@
...
+            WebhookEvent.PLAN_CHANGED,
+            webhook_payload,
+            WebhookStatus.IN_PROGRESS,
+        )
+        await self.send_notifications(logger)
+
+
+class AfterChangePlanSubHandler(StateBasedSubHandler):
+    """
+    A handler which depends on``restart`` having finishe successfully and sends a
```
SStorm, line 39:
```suggestion
    A handler which depends on``restart`` having finished successfully and sends a
```

crate/operator/change_plan.py
```diff
@@ -0,0 +1,121 @@
...
+                        {
+                            "name": "crate",
+                            "resources": {
+                                "limits": {
+                                    "cpu": plan_change_data["new_cpu_limit"],
+                                    "memory": plan_change_data["new_memory_limit"],
+                                },
+                                "requests": {
+                                    "cpu": plan_change_data["new_cpu_request"]
+                                    or plan_change_data["new_cpu_limit"],
```
SStorm, line 101: What happens if `new_cpu_request`  is not set - does `plan_change_data` have it but it's `None`, or is the key not in the dictionary at all? In the latter case you would get an exception here...

crate/operator/handlers/handle_update_cratedb.py
```diff
@@ -119,8 +120,28 @@ async def update_cratedb(
                     "size"
                 ) != new_spec.get("resources", {}).get("disk", {}).get("size"):
                     do_expand_volume = True
-
-    if not do_upgrade and not do_restart and not do_scale and not do_expand_volume:
+                elif (
```
SStorm, line 123: Can we perhaps simplify this `if` with some sort of helper function? Sure we could write some fancy for loop here :)

tests/test_change_plan.py
```diff
@@ -0,0 +1,159 @@
...
+        group=API_GROUP,
+        version="v1",
+        plural=RESOURCE_CRATEDB,
+        namespace=namespace.metadata.name,
+        name=name,
+        body=[
+            {
+                "op": "replace",
+                "path": "/spec/nodes/data/0/resources/limits/cpu",
+                "value": 1,  # Increase from 2
```
SStorm, line 93:
```suggestion
                "value": 1,  # Decrease from 2
```

tests/test_change_plan.py
```diff
@@ -0,0 +1,159 @@
...
+
+    await assert_wait_for(
+        False,
+        do_pod_ids_exist,
+        core,
+        namespace.metadata.name,
+        original_pods,
+        timeout=DEFAULT_TIMEOUT * 15,
+    )
+
```
SStorm, line 131: You might also want to assert that the limits/requests have actually changed on the new pods.

crate/operator/change_plan.py
SStorm: Also need to align with @Taliik what the expected webhooks should be...

crate/operator/change_compute.py
```diff
@@ -0,0 +1,157 @@
...
+                    plan_change_data["new_memory_limit"],
+                ),
+            },
+        },
+    }
+    body = {
+        "spec": {
+            "template": {
+                "spec": {
+                    "affinity": get_statefulset_affinity(name, logger, node_spec),
```
SStorm, line 137: Do we also need to update the affinity when updating the containers spec?

tests/test_change_compute.py
```diff
@@ -0,0 +1,318 @@
...
+    assert_wait_for,
+    cluster_routing_allocation_enable_equals,
+    create_test_sys_jobs_table,
+    is_cluster_healthy,
+    is_kopf_handler_finished,
+    start_cluster,
+)
+
+
+async def do_pods_exist(core: CoreV1Api, namespace: str, expected: Set[str]) -> bool:
```
SStorm, line 39: I believe these methods exist already - maybe worth moving them to `utils.py`

crate/operator/handlers/handle_update_cratedb.py
```diff
@@ -119,8 +124,18 @@ async def update_cratedb(
...
+                    do_change_compute = True
+                    # pod resources won't change until each pod is recreated
+                    do_restart = True
+
+    if (
+        not do_upgrade
+        and not do_restart
+        and not do_scale
+        and not do_expand_volume
+        and not do_change_compute
```
SStorm, line 137: There's a warning here about usage-before-assign - I think you might need to initialize it above around line 99.

tests/test_change_compute.py
```diff
@@ -0,0 +1,318 @@
...
+        )
+
+    coapi = CustomObjectsApi(api_client)
+    core = CoreV1Api(api_client)
+    name = faker.domain_word()
+    if set_initial_request_resources:
+        # Start a cluster with requests set to half the original limits
+        initial_requests = {"cpu": 1, "memory": "2Gi"}
+        host, password = await start_cluster(
+            name, namespace, core, coapi, 3, "5.0.0", resource_requests=initial_requests
```
SStorm, line 128: Would be better to only start a 1 node cluster here - it takes wayyyy longer for 3 nodes to mount PVCs/initialize/restart etc.

tests/test_change_compute.py
```diff
@@ -0,0 +1,318 @@
...
+    pods = await core.list_namespaced_pod(namespace=namespace)
+    return expected.issubset({p.metadata.name for p in pods.items})
+
+
+async def do_pod_ids_exist(core: CoreV1Api, namespace: str, pod_ids: Set[str]) -> bool:
+    pods = await core.list_namespaced_pod(namespace=namespace)
+    return bool(pod_ids.intersection({p.metadata.uid for p in pods.items}))
+
+
+@pytest.mark.k8s
```
SStorm, line 49: While it's great that we test multiple possible cases, these tests take circa 2 minutes each and add _significantly_ to the overall test duration... Perhaps worth testing just one of the scenarios? WDYT?

tests/test_change_compute.py
```diff
@@ -0,0 +1,305 @@
...
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "set_initial_request_resources, cpu_request, memory_request, cpu_limit,"
+    "memory_limit",
+    [
+        # Test a change in cpu and memory from req != limits to req = limits
+        (True, 1, "5Gi", 1, "5Gi"),
```
SStorm, line 47: You don't need the params any more now that there's just one :)

SStorm (approved):
:100:

tests/test_change_compute.py
```diff
@@ -0,0 +1,305 @@
+# CrateDB Kubernetes Operator
```
SStorm, line 1: Could you please change the license to Apache as other files, thanks :)

SStorm (approved):
:100:

PR #392, opened 2022-09-02
create Helm chart

plaharanne:
## Summary of changes

https://github.com/crate/cloud/issues/614

## Checklist


WalBeh:
LGTM, maybe consider adding a `NOTES.txt`? https://helm.sh/docs/chart_template_guide/notes_files/

deploy/charts/crate-operator/values.yaml
```diff
@@ -0,0 +1,68 @@
...
+  # If not set and create is true, a name is generated using the fullname template
+  name: "crate-operator"
+
+rbac:
+  # Specifies whether RBAC resources should be created
+  create: true
+
+podAnnotations: {}
+
+resources: {}
```
WalBeh, line 56: I guess it makes sense to set sensible defaults here, as _we_ should know.

deploy/charts/crate-operator/README.md
```diff
@@ -0,0 +1,57 @@
...
+The namespace needs either to be created first:
+
+```shell
+kubectl create namespace crate-operator
+```
+
+or it will be created automatically by adding `--namespace=crate-operator --create-namespace` to the Helm command:
+
+```shell
+helm install crate-operator crate/crate-operator \
```
WalBeh, line 50:
```suggestion
helm upgrade --install crate-operator crate/crate-operator \
```

deploy/charts/crate-operator-crds/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v2
+name: crate-operator-crds
+description: Crate Operator CRDs - Helm chart for installing and upgrading Custom Resource Definitions (CRDs) for the Crate Operator.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
```
SStorm, line 9: What do you think about the version here - should we keep it in sync with the operator version? What is the best practice?

deploy/charts/crate-operator-crds/README.md
```diff
@@ -0,0 +1,26 @@
...
+To be able to deploy the custom resource CrateDB to a Kubernetes cluster, the API needs to be extended with a Custom Resource Definition (CRD).
+
+Helm must be installed to use the charts. Please refer to Helm's [documentation](https://helm.sh/docs/) to get started.
+
+## Usage
+
+Once Helm is properly set up, add the repo:
+
+```console
+helm repo add crate https://
```
SStorm, line 13: Should we document installing from the local folder?

deploy/charts/crate-operator/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+name: crate-operator
+description: Crate Operator - Helm chart for installing and upgrading Crate Operator.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
```
SStorm, line 9: Same question about the version :thinking:

deploy/charts/crate-operator/templates/rbac.yaml
```diff
@@ -0,0 +1,72 @@
+{{- if .Values.rbac.create -}}
```
SStorm, line 1: Do we need to have an `if` for the rbac? When would we not want to install this?

deploy/charts/crate-operator/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.serviceAccount.create -}}
```
SStorm, line 1: And same for this...

deploy/charts/crate-operator/values.yaml
SStorm: Agree, would be nice to set some sensible defaults.

SStorm (commented):
Nice work Pauline - I've added a few questions below - happy to discuss!

deploy/charts/crate-operator-crds/README.md
plaharanne: I waited to have set up the github action to publish the chart before competing the repo url here, it should be `https://crate.github.io/crate-operator` in theory.
I would rather document using the repo instead of the local folder as the local method is supposed to me more for testing and stuff like that.

deploy/charts/crate-operator/templates/rbac.yaml
plaharanne: As the ClusterRole and ClusterRoleBinding are not linked to a specific namespace, I would rather let the user the possibility to disable it.
I don't think it should happen often, or maybe never, but I'd like better to have the possibility in case it happens instead of not being able to continue the installation.

deploy/charts/crate-operator/templates/serviceaccount.yaml
plaharanne: I used `helm create <name>` to create the Chart structure and I didn't really change the template for the serviceAccount.
I can change it if you think it's a problem.

deploy/charts/crate-operator/values.yaml
```diff
@@ -0,0 +1,68 @@
...
+  # the CRD dependency automatically.
+  enabled: true
+
+replicaCount: 1
+
+image:
+  repository: crate/crate-operator
+  pullPolicy: Always
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: latest
```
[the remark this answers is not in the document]
plaharanne: In the deployment script (jenkins) it is overwritten ( `--set image.tag=${APP_VERSION}` )
I don't think it is a problem to let the default value as `latest` otherwise we will have to maintain it here as well.

deploy/charts/crate-operator/values.yaml
plaharanne: I agree with you but I'm not sure which one.

deploy/charts/crate-operator/values.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+imagePullSecrets: []
+nameOverride: "crate-operator"
+fullnameOverride: "crate-operator"
+
+envVars:
+  - name: CRATEDB_OPERATOR_BOOTSTRAP_TIMEOUT
```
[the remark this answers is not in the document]
plaharanne: I put a link to the [doc](https://github.com/crate/crate-operator/blob/master/docs/source/configuration.rst) in the readme for that.

deploy/charts/crate-operator-crds/Chart.yaml
plaharanne: idk what to think about it. If it's the same, it's easier to figure out which release is the good one. On the other side, if we want to change the chart or its templates we would have to change the operator version too.

SStorm: I would be ok with that, would rather the charts and the operator are a single distribution that share a version. But not sure what the best practice is. WDYT @WalBeh ?

deploy/charts/crate-operator/templates/serviceaccount.yaml
SStorm: I am ok with it if this is what helm recommends, just curious why.

deploy/charts/crate-operator/values.yaml
SStorm: Let's set requests/limits to `250m` CPU and `128Mi` - just tried it on dev and it seems ok.

deploy/charts/crate-operator/values.yaml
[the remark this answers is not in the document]
SStorm: btw, maybe worth having this commented out then? We already need to maintain the chart `AppVersion`, and if the tag is not specified - it will default to that :thinking:

deploy/charts/crate-operator/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "2.13.3"
```
SStorm, line 15: Let's also update the release procedure to change this when we release a new version of the operator?

SStorm (approved):
Looks good, thanks Pauline!

As discussed, we still need to write a script that helps us making releases, and update DEVELOP.rst with a new release process.

PR #398, opened 2022-09-15
Added support for parallel testing

SStorm:
## Summary of changes

This is done using pytest-xdist and by adding the `-n X` parameter to pytest, where X is the number of parallel executors.

This required a locking mechanism to be implemented, as each worker spawns it's own operator (in the `kopf_runner` fixture). If the tests detect that they are running in parallel mode, the operator will only process the CrateDBs that have the `{"testing": "$PID"}` annotation - meaning that they will only process CrateDBs from the same worker. This allows us to continue using various `@mock.patch` monkey-patches that assume there's just one process running.

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changelog
 Unreleased
 ----------
 
+* Added support for parallel testing, which greatly reduces test runtime.
+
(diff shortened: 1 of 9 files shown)


tomach (approved):
very cool :100:

PR #401, opened 2022-09-21
PoC to switch from the all-encompassing update handler

SStorm:
## Summary of changes

... to individual field handlers. This has been done for the upgrade operation only.

Biggest issue by far is that the semantics of the new and old fields changes - they now refer not to the entire body, but to the individual field that changed.

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -146,7 +146,7 @@ def cratedb_crd(request, tmp_path_factory, load_config):
     with FileLock(f"{root_tmp_dir}/operator-apply-lock"):
         subprocess.run(
             [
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -146,7 +146,7 @@ def cratedb_crd(request, tmp_path_factory, load_config):
     with FileLock(f"{root_tmp_dir}/operator-apply-lock"):
         subprocess.run(
             [
-                "kubectl",
+                "/home/romanas/bin/kubectl",
```
SStorm, line 149:
```suggestion
                "kubectl",
```

PR #410, opened 2022-10-07
Added cratedb_cluster_last_activity metric to the sql exporter

plaharanne:
## Summary of changes

https://github.com/crate/cloud/issues/589

## Checklist

### DIFF
--- a/crate/operator/data/cratedb_cluster_last_activity-collector.yaml
+++ b/crate/operator/data/cratedb_cluster_last_activity-collector.yaml
@@ -1,0 +1,14 @@
+---
+collector_name: cratedb_cluster_last_activity_collector
+# This metric is intended to know if the cluster is used
+# and to calculate for how long it has not been used.
+metrics:
+- help: "Indicates when the last job occured on the cluster."
+  metric_name: cratedb_cluster_last_activity
+  query: |
+    SELECT max(ended) AS cratedb_cluster_last_activity
+    FROM sys.jobs_log
+    WHERE username NOT IN ('crate', 'system');
+  type: gauge
+  value_label: cratedb_cluster_last_activity
+  values: [cratedb_cluster_last_activity]
--- a/crate/operator/data/cratedb_cluster_last_user_activity-collector.yaml
+++ b/crate/operator/data/cratedb_cluster_last_user_activity-collector.yaml
@@ -6,7 +6,7 @@ metrics:
 - help: "Indicates when the last job occured on the cluster."
   metric_name: cratedb_cluster_last_user_activity
   query: |
-    SELECT max(ended) AS cratedb_cluster_last_user_activity
+    SELECT cast(extract(epoch from max(ended)) as integer) AS cratedb_cluster_last_user_activity
     FROM sys.jobs_log
     WHERE username NOT IN ('crate', 'system');
   type: gauge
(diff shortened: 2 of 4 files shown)


crate/operator/data/cratedb_cluster_last_activity-collector.yaml
```diff
@@ -0,0 +1,14 @@
+---
+collector_name: cratedb_cluster_last_activity_collector
```
SStorm, line 2: I wonder if we should call it `cratedb_cluster_last_user_activity` to clearly state that this is only for user-initiated stuff and not system stuff...

SStorm (approved):
Looks good, but let's make sure this works as a metric, i.e. it's correctly displayed in prometheus etc.

crate/operator/data/cratedb_cluster_last_user_activity-collector.yaml
```diff
@@ -0,0 +1,14 @@
...
+# This metric is intended to know if the cluster is used
+# and to calculate for how long it has not been used.
+metrics:
+- help: "Indicates when the last job occured on the cluster."
+  metric_name: cratedb_cluster_last_user_activity
+  query: |
+    SELECT max(ended) AS cratedb_cluster_last_user_activity
+    FROM sys.jobs_log
+    WHERE username NOT IN ('crate', 'system');
+  type: gauge
```
WalBeh, line 12: Not sure, but maybe use a `counter` https://prometheus.io/docs/concepts/metric_types/ as this is expected to increase over time, unless we travel in time.

crate/operator/data/sql-exporter.yaml
```diff
@@ -14,7 +14,7 @@ global:
   # timing out first.
   scrape_timeout_offset: 500ms
```
WalBeh, line 15: As far as I see this is the `default`.

crate/operator/data/cratedb_cluster_last_user_activity-collector.yaml
```diff
@@ -0,0 +1,14 @@
...
+  value_label: cratedb_cluster_last_user_activity
+  values: [cratedb_cluster_last_user_activity]
```
WalBeh, line 14: No string opinion here, but maybe use something like `cratedb_clusters_user_last_seen` (I think most other metrics use `clusters`.

crate/operator/data/cratedb_cluster_last_user_activity-collector.yaml
SStorm: It's a bit of a black magic, but the [recommendation](https://www.robustperception.io/are-increasing-timestamps-counters-or-gauges/) is to use gauges for time

crate/operator/data/cratedb_cluster_last_user_activity-collector.yaml
SStorm: I kind of like activity as it's more specific, in this case. Indicates that the user has actually done something, rather than was just "seen" (i.e. logged in)

plaharanne: We already have a "last_seen" column in the cluster table, it might be confusing to use `last_seen" for that too.

PR #421, opened 2022-10-27
Disabled parallel cluster suspension and volume resizing

SStorm:
## Summary of changes

This was causing issues on Azure AKS. Will now first suspend the cluster and only then attempt to resize volumes.

## Checklist

### DIFF
--- a/crate/operator/handlers/handle_update_cratedb.py
+++ b/crate/operator/handlers/handle_update_cratedb.py
@@ -207,21 +207,15 @@ async def update_cratedb(
 def register_storage_expansion_handlers(
     namespace: str, name: str, change_hash: str, context: dict, depends_on: list
 ):
-    # Volume expansion and cluster suspension can run in parallel. Scaling the
-    # cluster back up needs to wait until both operations are finished.
-    kopf.register(
-        fn=ExpandVolumeSubHandler(
-            namespace, name, change_hash, context, depends_on=depends_on.copy()
-        )(),
-        id="expand_volume",
-        backoff=get_backoff(),
-    )
     if config.NO_DOWNTIME_STORAGE_EXPANSION:
-        depends_on.extend(
-            [
-                f"{CLUSTER_UPDATE_ID}/expand_volume",
-            ]
+        kopf.register(
+            fn=ExpandVolumeSubHandler(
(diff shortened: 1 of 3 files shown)


crate/operator/handlers/handle_update_cratedb.py
```diff
@@ -207,21 +207,15 @@ async def update_cratedb(
...
-        )(),
-        id="expand_volume",
-        backoff=get_backoff(),
-    )
     if config.NO_DOWNTIME_STORAGE_EXPANSION:
-        depends_on.extend(
-            [
-                f"{CLUSTER_UPDATE_ID}/expand_volume",
-            ]
+        kopf.register(
```
juanpardo, line 211: I think the registration of ``ExpandVolumeSubhandler`` can go outside of the if-else clause, since it's in both sides.
You probably want to suspend first if ``config.NO_DOWNTIME_STORAGE_EXPANSION`` is false, then call the ``ExpandVolumeSubhandler`` regardless

SStorm: I think that makes it a bit messy, as I also need to conditionally include the `StartClusterSubHandler` in that case. So would be something like:

```python
if not config.NO_DOWNTIME_STORAGE_EXPANSION:
  suspend_sub_handler()

expand_sub_handler()

if not config.NO_DOWNTIME_STORAGE_EXPANSION:
  start_sub_handler()
```

I find it easier to read if there are two distinct branches that handle it in the way they need to.

juanpardo: In that case please encapsulate the repeated logic in a function

SStorm: So you want to move:

```python
kopf.register(
  fn=ExpandVolumeSubHandler(
      namespace, name, change_hash, context, depends_on=depends_on.copy()
  )(),
  id="expand_volume",
  backoff=get_backoff(),
)
depends_on.append(f"{CLUSTER_UPDATE_ID}/expand_volume")
```

into a separate function? We don't do that for the other handler calls, so I don't see a particularly clear benefit to it. Would just be a single dangling function.

juanpardo (approved):

PR #425, opened 2022-11-02
Update backup cronjob schedule

juanpardo:
## Summary of changes
Cluster backup cronjob schedules can now be updated.

## Checklist


crate/operator/handlers/handle_update_cratedb.py
```diff
@@ -138,13 +140,16 @@ async def update_cratedb(
                     do_change_compute = True
                     # pod resources won't change until each pod is recreated
                     do_restart = True
+        elif field_path == ("spec", "backups", "aws", "cron"):
```
SStorm, line 143: I believe we discussed perhaps using a field handler (i.e. similar to CIDR changes) for this instead of adding more stuff to the update handler? We don't really need to do any of the other stuff that the update handler handles for cronjob changes (i.e. disabling backups).

crate/operator/change_backup.py
```diff
@@ -0,0 +1,57 @@
...
+        for d in diff:
+            if d.operation == "change" and d.field == (
+                "spec",
+                "backups",
+                "aws",
+                "cron",
+            ):
+                diff_found = d
+
+        assert diff_found is not None
```
SStorm, line 41: If this happens, the handler would re-run forever (I think?). If we want to stop it completely, we need to raise a `kopf.PermanentError()`

SStorm (commented):
I think it would be better to do this as a field handler, instead of adding more stuff to the already pretty bloated update handler, i.e. as done [here](https://github.com/crate/crate-operator/blob/master/crate/operator/main.py#L180). 

I don't think we need to do the pre- and post-update things when we simply change the CronJob schedule - same as the CIDR or password change which are done in field handlers.

crate/operator/main.py
```diff
@@ -215,6 +216,26 @@ async def service_external_ip_update(
...
+)
+async def service_backup_schedule_update(
+    namespace: str,
+    name: str,
+    diff: kopf.Diff,
+    logger: logging.Logger,
+    **_kwargs,
+):
+    """
+    Handles updates to the list of allowed CIDRs, and updates the relevant k8s Service.
```
SStorm, line 234:
```suggestion
    Handles updates to the backup schedule for AWS s3 backups.
```

tests/test_update_backups_schedule.py
```diff
@@ -0,0 +1,141 @@
...
+    await assert_wait_for(
+        True,
+        is_cluster_healthy,
+        conn_factory,
+        1,
+        err_msg="Cluster wasn't healthy",
+        timeout=DEFAULT_TIMEOUT,
+    )
+
+    await create_test_sys_jobs_table(conn_factory)
```
SStorm, line 90: You don't need the sys_jobs table, unless testing delays while backups are running.

SStorm (approved):
💯

crate/operator/handlers/handle_update_cratedb.py
juanpardo: Changed it to use the field handler

crate/operator/change_backup.py
juanpardo: This doesn't make sense anymore because with the field handler it is guaranteed to only have 1 field

tests/test_update_backups_schedule.py
juanpardo: It was failing without it before, but it was due to beforeUpdate handler, which is not triggered anymore

PR #432, opened 2022-11-28
Add support for restoring clusters from a backup

tomach:
## Summary of changes
This change allows restoring a backup from a different source crateDB cluster into another crateDB cluster. To trigger this operation, an object called `spec.cluster.restoreSnapshot` has to be created in the cratedb crd e.g. like this 

```yml
    restoreSnapshot:
      accessKeyId:
        secretKeyRef:
          key: access-key-id
          name: restore-from-backup-{target_cluster_id}
      basePath:
        secretKeyRef:
          key: base-path
          name: restore-from-backup-{target_cluster_id}
      bucket:
        secretKeyRef:
          key: bucket
          name: restore-from-backup-{target_cluster_id}
      secretAccessKey:
        secretKeyRef:
          key: secret-access-key
          name: restore-from-backup-{target_cluster_id}
      snapshot: "{snapshot_name}"
      tables:
      - all
```

Also the secret called `restore-from-backup-{target_cluster_id}` in the target cluster's namespace containing `bucket`, `base-path`, `secret-access-key`, `access-key-id` has to be created as specified above. Note that the secrets could be read from any other secret, but only secrets named `restore-from-backup-{cluster_id}` are automtaically deleted at the end of the restore operation for safety reasons.

The operator handles changes to the field `spec.cluster.restoreSnapshot`. In each iteration it checks at the beginning if there is a restore operation already in progress by querying the `sys.snapshot_restore` table. If there is a restore in progress, it only reports a status update to the API. Nothing else is done while a restore is in progress.

The operation consists basically of 3 parts:
- `before_restore_handlers`
  - `BeforeClusterUpdateSubHandler` - this one is already known from other cluster operations. It ensures there is no snapshot in progress, disables cronjobs, etc.
  - `BeforeRestoreBackupSubHandler` - performs restore specific preparation tasks. It suspends the backup-metrics deployment and updates cluster settings `cluster.routing.allocation.cluster_concurrent_rebalance` and `indices.recovery.max_bytes_per_sec` (it preserves the old values in the `status` object of the crd to be able to reset them after the restore)
- `restore_handlers`
  - `RestoreBackupSubHandler` - the main operation happens here.
    - it reads the necessary credentials from the given secret to be able to access the source backup repository
    - it creates a temporary backup repository with the given secrets
    - it ensures that the given snapshot name exists in the repository
    - it performs a `RESTORE SNAPSHOT` with the given table names or `all`
    (Note: most errors in this part raise a `PermanentError` because I think it is unlikely that retries would help if a snapshot is not available in a repository, a secret contains invalid credentials etc.)
  - `RestoreSystemUserPasswordSubHandler` - it reads the system user's password from the secret and restores it using `crash` (during the restore operation the `sys.users` table was probably overwritten)
- `after_restore_handlers`
  - `AfterRestoreBackupSubHandler` - cleanup after restore
    - it resets the cluster settings `cluster.routing.allocation.cluster_concurrent_rebalance` and `indices.recovery.max_bytes_per_sec` to the preserved values
    - it drops the temporary backup repository
    - it deletes the secret containing the credentials (only if it is named `restore-from-backup-{cluster_id}`)
    - it scales the backup-metrics deployment back up
  - `AfterClusterUpdateSubHandler` - already known from other cluster operations. It re-enables the cronjob etc.

In the end, a success notification is sent to the API and the value of `spec.cluster.restoreSnapshot.snapshot` is deleted to be able to restore the same snapshot again in case anything goes wrong.

https://github.com/crate/cloud/issues/546

## Checklist

### DIFF
--- a/crate/operator/handlers/handle_restore_backup.py
+++ b/crate/operator/handlers/handle_restore_backup.py
@@ -1,0 +1,271 @@
+# CrateDB Kubernetes Operator
+#
+# Licensed to Crate.IO GmbH ("Crate") under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  Crate licenses
+# this file to you under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.  You may
+# obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+# However, if you have executed another commercial license agreement
+# with Crate these terms will supersede the license and you may use the
+# software solely pursuant to the terms of the relevant commercial agreement.
+
+import datetime
+import hashlib
+import logging
+from typing import List
+
+import kopf
+
+from crate.operator.config import config
+from crate.operator.handlers.handle_update_cratedb import get_backoff
+from crate.operator.operations import (
+    AfterClusterUpdateSubHandler,
+    BeforeClusterUpdateSubHandler,
+    get_cratedb_resource,
+)
+from crate.operator.restore_backup import (
+    AfterRestoreBackupSubHandler,
+    BeforeRestoreBackupSubHandler,
+    ResetSnapshotSubHandler,
+    RestoreBackupSubHandler,
+    RestoreSystemUserPasswordSubHandler,
+    SendSuccessNotificationSubHandler,
+    ensure_no_restore_in_progress,
+    get_crash_pod_name,
+    get_crash_scheme,
+)
+from crate.operator.utils.notifications import FlushNotificationsSubHandler
+
+CLUSTER_RESTORE_FIELD_ID = "cluster_restore/spec.cluster.restoreSnapshot"
+
+
+async def restore_backup(
+    namespace: str,
+    name: str,
+    diff: kopf.Diff,
+    new: kopf.Body,
+    patch: kopf.Patch,
+    status: kopf.Status,
+    started: datetime.datetime,
+    logger: logging.Logger,
+):
+    context = status.get(CLUSTER_RESTORE_FIELD_ID)
+    hash_string = str(diff) + str(started)
+    change_hash = hashlib.md5(hash_string.encode("utf-8")).hexdigest()
+    if not context:
+        context = {"ref": change_hash}
+    elif context.get("ref", "") != change_hash:
+        context["ref"] = change_hash
+
+    cratedb = await get_cratedb_resource(namespace, name)
+    snapshot = new["snapshot"]
+    tables = new.get("tables", ["all"])
+
+    scheme = get_crash_scheme(cratedb)
+    pod_name = get_crash_pod_name(cratedb, name)
+    repository = f"restore_backup_{int(started.timestamp())}"
+
+    # If there is a restore operation in progress, we do not do
+    # anything else until it is finished.
+    await ensure_no_restore_in_progress(
+        namespace, name, snapshot, pod_name, scheme, logger
+    )
+
+    depends_on: List[str] = []
+
+    register_before_restore_handlers(
+        namespace, name, patch, change_hash, context, depends_on
+    )
+
+    register_restore_handlers(
+        namespace,
+        name,
+        change_hash,
+        context,
+        depends_on,
+        repository,
+        snapshot,
+        tables,
+    )
+
+    register_after_restore_handlers(
+        namespace,
+        name,
+        status,
+        change_hash,
+        context,
+        depends_on,
+        repository,
+    )
+
+    kopf.register(
+        fn=FlushNotificationsSubHandler(
+            namespace,
+            name,
+            change_hash,
+            context,
+            depends_on=depends_on.copy(),
+            run_on_dep_failures=True,
+        )(),
+        id="flush_notifications",
+        backoff=get_backoff(),
+    )
+    depends_on.append(f"{CLUSTER_RESTORE_FIELD_ID}/flush_notifications")
+
+    # This needs to be the last subhandler, otherwise the operation does
+    # not finish properly.
+    kopf.register(
+        fn=ResetSnapshotSubHandler(
+            namespace,
+            name,
+            change_hash,
+            context,
+            depends_on=depends_on.copy(),
+            run_on_dep_failures=True,
+        )(),
+        id="reset_snapshot",
+        backoff=get_backoff(),
+    )
+
+    patch.status[CLUSTER_RESTORE_FIELD_ID] = context
+
+
+def register_before_restore_handlers(
+    namespace: str,
+    name: str,
+    patch: kopf.Patch,
+    change_hash: str,
+    context: dict,
+    depends_on: list,
+):
+
+    kopf.register(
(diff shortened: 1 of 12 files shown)


crate/operator/restore_backup.py
```diff
@@ -0,0 +1,925 @@
...
+                                "base_path = %s, readonly=true)",
+                                (
+                                    data["accessKeyId"],
+                                    data["secretAccessKey"],
+                                    data["bucket"],
+                                    data["basePath"],
+                                ),
+                            )
+                        except ProgrammingError as e:
+                            logger.warning(
```
SStorm, line 549: When can this happen? Should this also raise `PermanentError`? Why not add it to the exception handler below?

crate/operator/restore_backup.py
```diff
@@ -0,0 +1,925 @@
...
+                    repository_ident = quote_ident(repository, cursor._impl)
+                    snapshot_ident = quote_ident(snapshot, cursor._impl)
+                    try:
+                        await cursor.execute(
+                            f"RESTORE SNAPSHOT {repository_ident}.{snapshot_ident} "
+                            f"{AsIs(tables_str)} with (wait_for_completion=false)"
+                        )
+                    except DuplicateTable as e:
+                        logger.warning("Relation already exists.", exc_info=e)
+                        raise kopf.PermanentError(
```
SStorm, line 641: Does this happen if any table from the snapshot already exists? What happens to the other tables in that case?..

crate/operator/restore_backup.py
```diff
@@ -0,0 +1,925 @@
...
+                            "ProgrammingError in _start_restore_snapshot.", delay=15
+                        )
+        except (DatabaseError, OperationalError) as e:
+            logger.warning("DatabaseError in _start_restore_snapshot: %s", str(e))
+            if "index with same name already exists" in str(e):
+                raise kopf.PermanentError("Index with the same name already exists.")
+            elif "mismatched input" in str(e) or "template not found" in str(e):
+                raise kopf.PermanentError("Index not found in given snapshot.")
+            else:
+                raise kopf.TemporaryError(
```
SStorm, line 656: Also same thought - can we recover from this?

crate/operator/restore_backup.py
```diff
@@ -0,0 +1,925 @@
...
+            (schema, table_name),
+        )
+        row = await cursor.fetchone()
+        if row:
+            logger.info("found shard with id %s ", row[0])
+            in_progress.append(int(row[0]))
+    return in_progress
+
+
+async def is_restore_complete(
```
SStorm, line 191: Doesn't the `ensure_no_restore_in_progress` ensures that a snapshot has completed already?

SStorm (approved):
Well done @tomach 💯 

I've added a bunch of questions but approving this anyway as it seems legit.

crate/operator/restore_backup.py
```diff
@@ -0,0 +1,925 @@
...
+def is_valid_snapshot(new: kopf.Body, **kwargs) -> bool:
+    """
+    This checks if the new snapshot name is valid or not (empty).
+    This check is necessary to avoid another restore operation is
+    triggered after we reset the snapshot field at the end of a
+    completed restore operation.
+
+    :param new: The new CrateDB resource
+    """
+    return len(new["spec"]["cluster"]["restoreSnapshot"]["snapshot"]) > 0
```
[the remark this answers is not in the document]
tomach: Hm true, if the whole `restoreSnapshot` object was deleted, it raises an error, will fix it. Only `snapshot` cannot be `None` I think since it is defined as `required` and `type: string`

crate/operator/restore_backup.py
tomach: true, this actually inherits from `DatabaseError` and can be handled below..

crate/operator/handlers/handle_restore_backup.py
```diff
@@ -0,0 +1,273 @@
...
+            context,
+            depends_on=depends_on.copy(),
+            run_on_dep_failures=True,
+        )(),
+        id="flush_notifications",
+        backoff=get_backoff(),
+    )
+    depends_on.append(f"{CLUSTER_RESTORE_FIELD_ID}/flush_notifications")
+
+    # This needs to be the last subhandler, otherwise the operation does
```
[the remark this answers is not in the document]
tomach: Like this, it seems to finish properly. All the annotations with the subhandler status are deleted.

```
Handler 'cluster_restore/spec.cluster.restoreSnapshot' succeeded.
Updating is processed: 2 succeeded; 0 failed.
```

I think 1 for the initial update and the 2nd for the update to `snapshot: ""` which does not really run because `when=is_valid_snapshot` returns `False`.

crate/operator/restore_backup.py
```diff
@@ -0,0 +1,925 @@
...
+                                    data["basePath"],
+                                ),
+                            )
+                        except ProgrammingError as e:
+                            logger.warning(
+                                "CREATE REPOSITORY query failed.", exc_info=e
+                            )
+        except (DatabaseError, OperationalError) as e:
+            logger.warning("DatabaseError in _create_backup_repository: %s", str(e))
+            if "Unable to verify the repository" in str(e):
```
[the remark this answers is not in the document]
tomach: Unfortunately I didn't find another way.. `e.pgcode` is always `XX000` for `internal_error` and `e.pgerror` sth like
```
[restore_backup_12345] Unable to verify the repository, [restore_backup_12345] is not accessible on master node: AmazonS3Exception 'The request signature we calculated does not match the signature you provided. Check your key and signing method. (Service: Amazon S3; Status Code: 403; Error Code: SignatureDoesNotMatch; Request ID: XXX; S3 Extended Request ID: XXX; Proxy: null)
```
with lots of different `Error Code` possibilities from AWS, only `Unable to verify the repository` was always the same.
But maybe we should just raise a `PermanentError` without further checks, if any of these queries fails? :thinking:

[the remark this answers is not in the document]
SStorm: > But maybe we should just raise a PermanentError without further checks, if any of these queries fails? 🤔

I think so, because I can't imagine any cases where it could fail temporarily 🤔

crate/operator/restore_backup.py
tomach: yes, if at least one of the tables to restore already exists, the whole `RESTORE SNAPSHOT` operation fails.

crate/operator/restore_backup.py
```diff
@@ -0,0 +1,925 @@
...
+            "where state='RECOVERING' and recovery['type']='SNAPSHOT';"
+        )
+        result = await run_crash_command(
+            namespace, pod_name, scheme, progress_command, logger
+        )
+        pct = int(re.findall(r"(\d+)", result)[0]) or 0
+        await send_operation_progress_notification(
+            namespace=namespace,
+            name=name,
+            message=f"Please wait while cluster data is being restored... ({pct}%)",
```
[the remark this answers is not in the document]
tomach: It's the `min()` over all shards in progress. I had `avg()` before, but the number jumped up and down when the different shards finished :)

crate/operator/restore_backup.py
SStorm: Cool, did not know that! We might want to explain that in the UI...

crate/operator/restore_backup.py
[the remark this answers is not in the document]
tomach: ok, I raise a `PermanentError` now if anything goes wrong in one of the repository/snapshot queries and removed the checks for certain strings in the error msg.

crate/operator/restore_backup.py
tomach: true, I think this check is not necessary, I'll remove it..

PR #451, opened 2023-01-11
Cluster cloning restores admin username in CRD

juanpardo:
## Checklist

### DIFF
--- a/crate/operator/restore_backup.py
+++ b/crate/operator/restore_backup.py
@@ -38,6 +38,7 @@ from crate.operator.config import config
 from crate.operator.constants import API_GROUP, RESOURCE_CRATEDB, SYSTEM_USERNAME
 from crate.operator.cratedb import (
     connection_factory,
+    get_cluster_admin_username,
     get_cluster_settings,
     set_cluster_setting,
 )
@@ -55,8 +56,8 @@ from crate.operator.utils.kubeapi import (
 from crate.operator.utils.notifications import send_operation_progress_notification
 from crate.operator.webhooks import (
     WebhookEvent,
-    WebhookFeedbackPayload,
     WebhookOperation,
+    WebhookSnapshotRestoredPayload,
     WebhookStatus,
 )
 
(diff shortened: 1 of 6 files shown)


crate/operator/webhooks.py
```diff
@@ -37,6 +37,7 @@ class WebhookEvent(str, enum.Enum):
     HEALTH = "health"
     FEEDBACK = "feedback"
     BACKUP_SCHEDULE_UPDATE = "backup_schedule_update"
+    SNAPSHOT_RESTORED = "snapshot_restored"
```
SStorm, line 40: This also needs handling in the API, right? How did it work before?

crate/operator/restore_backup.py
```diff
@@ -574,6 +584,51 @@ async def handle(  # type: ignore
                 logger.info("... error. %s", result)
                 raise kopf.TemporaryError(delay=config.BOOTSTRAP_RETRY_DELAY)
 
+            # Determine if the admin username has changed.
+            command = (
+                "SELECT grantee FROM sys.privileges where grantor = 'system' limit 1"
+            )
+            result = await run_crash_command(
```
SStorm, line 591: Looks like you do not need this at all as you ignore the result?

SStorm (commented):
Can @tomach also take a look as he's got most experience with this feature ;)

crate/operator/webhooks.py
tomach: At the moment it works with the `FEEDBACK` event type only. This type only supports `operation` and `message`, so I think it's a good idea to introduce a new event type for reporting back the admin username. But I think we should not mix notifications for the status of the restore operation and the admin username. With [this notification](https://github.com/crate/crate-operator/pull/451/files#diff-71c803b5c41e32172f7ec518522689fc1da8582141d29c6c26c61c32af88f07eR604) and the API changes in the other PR, we would mark the whole restore operation as succeeded, right? Although `RestoreSystemUserPasswordSubHandler` runs even if the restore itself failed (`run_on_dep_failures=True`)

What if we keep the status updates (also in the API) as they are with type `FEEDBACK` and just move this new part with the admin user check and the new notification to `SendSuccessNotificationSubHandler`? Then it would only update the CRD and send the webhook if everything went well. So I mean `SendSuccessNotificationSubHandler` would then schedule 2 notifications, the existing status success and a new one with type `SNAPSHOT_RESTORED` and the API would only update the cluster's username for the new event.

juanpardo: It used to work with the feedback webhook, which just sends a string back

juanpardo: Thanks for the clarification @tomach , I think that's a great plan. Then perhaps ``SNAPSHOT_RESTORED`` is not a very descriptive name, how about we call it something more descriptive like ``ADMIN_USERNAME_CHANGED``?

tomach: :+1: sounds good

crate/operator/restore_backup.py
```diff
@@ -749,6 +780,32 @@ async def handle(  # type: ignore
         :param name: The CrateDB custom resource name defining the CrateDB cluster.
         :param logger: the logger on which we're logging
         """
+
+        # Cloning a cluster will result in the destruction of all target cluster users.
```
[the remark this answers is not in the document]
juanpardo: It was there before, but @tomach suggested to move it here, which makes sense because the ``RestoreSystemUserPasswordSubHandler`` gets executed even for errors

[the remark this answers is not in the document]
tomach: I thought the same at first, but the `RestoreSystemUserPasswordSubHandler` runs even if the restore failed, whereas the admin username update we only want to do if everything was successful, right?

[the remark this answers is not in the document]
SStorm: ok cool, understood.

PR #453, opened 2023-01-16
Fixed tests not passing after restore backup pw reset fix

juanpardo:
## Checklist

### DIFF
--- a/crate/operator/restore_backup.py
+++ b/crate/operator/restore_backup.py
@@ -793,7 +793,8 @@ class SendSuccessNotificationSubHandler(StateBasedSubHandler):
 
             # Retrieve admin username from the CrateDB CRD
             cratedb = await get_cratedb_resource(namespace, name)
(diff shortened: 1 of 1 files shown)


crate/operator/restore_backup.py
```diff
@@ -793,7 +793,8 @@ async def handle(  # type: ignore
 
             # Retrieve admin username from the CrateDB CRD
             cratedb = await get_cratedb_resource(namespace, name)
-            crd_username = cratedb["spec"]["users"][0]["name"]
```
SStorm, line 796: Why did this stop working? :thinking:

tomach: It just made the test fail, because the test cluster used in the restore-backup test does not have any users defined when created (it did not need any for what it was testing till now). Should never happen on a real cluster.

SStorm (approved):

PR #452, opened 2023-01-13
Add check if all shards have been fully restored

tomach:
## Summary of changes
Check if all shards of all tables that were restored during a `restore snapshot` operation have been restored completely. If not, this indicates that the cluster ran out of disk space while restoring.

## Checklist

- [x] Relevant changes are reflected in `CHANGES.rst`
- [ ] Added or changed code is covered by tests
- [ ] Documentation has been updated if necessary
- [ ] Changed code does not contain any breaking changes (or this is a major version change)

PR #456, opened 2023-01-18
Add check if shards exist for all restored tables

tomach:
## Summary of changes
Follow-up to https://github.com/crate/crate-operator/pull/452

There seems to be another possible state for `sys.shards` when the cluster ran out of disk space during a restore snapshot operation. shards of tables that have not even started to be restored seem to be there with `primary = FALSE` and `state = UNASSIGNED`. So we have 2 checks now:
- (the new) check if there is at least 1 shard with `primary = TRUE` for each table that should have been restored.
- (the already existing) check for primary shards in progress having `recovery['size']['percent'] < 100`

## Checklist


SStorm (approved):
:+1:

PR #464, opened 2023-02-06
Improve cluster healthiness query

WalBeh:
## Summary of changes
Add `sys.node_checks` to the health query

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changelog
 Unreleased
 ----------
 
+* Include ``sys.node_checks`` for checking cluster healthiness.
+
 2.22.0 (2023-01-31)
 -------------------
 
--- a/crate/operator/cratedb.py
+++ b/crate/operator/cratedb.py
@@ -126,7 +126,13 @@ async def get_healthiness(cursor: Cursor) -> int:
 
     :param cursor: A database cursor to a current and open database connection.
     """
-    await cursor.execute("SELECT MAX(severity) FROM sys.health")
+    await cursor.execute(
+        "SELECT max(severity) FROM ("
+        "  SELECT MAX(severity) as severity FROM sys.health"
+        "  UNION"
+        '  SELECT 3 as "severity" FROM sys.cluster where master_node is null'
+        ") sub;"
+    )
     row = await cursor.fetchone()
     return row and row[0]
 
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -26,6 +26,8 @@ Unreleased
 2.20.0 (2022-12-15)
 -------------------
 
+* Include ``sys.node_checks`` for checking cluster healthiness.
```
plaharanne, line 29: you probably should move it to `unreleased`

juanpardo:
As I pointed out in the chat, there are some entries in the table that may trigger false alerts. 
For example, this one:
```The amount of shards on the node reached 90 % of the limit of the cluster setting `cluster.max_shards_per_node`. Creating new tables or partitions which would push the number of shards beyond 100 % of the limit will be rejected. https://cr8.is/d-node-check-8```
Has severity 2, but the cluster should remain GREEN, in my opinion. I would suggest doing ``max(severity) - 1`` for the ``node_checks`` table

juanpardo (approved):

WalBeh:
> As I pointed out in the chat, there are some entries in the table that may trigger false alerts. For example, this one: `` The amount of shards on the node reached 90 % of the limit of the cluster setting `cluster.max_shards_per_node`. Creating new tables or partitions which would push the number of shards beyond 100 % of the limit will be rejected. https://cr8.is/d-node-check-8 `` Has severity 2, but the cluster should remain GREEN, in my opinion. I would suggest doing `max(severity) - 1` for the `node_checks` table

@juanpardo Thank you for taking time to look at this! 

IMHO the `severity` setting of each of the _checks_ seems to be sensible. Eg. _check-8_ comes with severity `2`, which is fine - I assume. If a check with severity `3` does not _pass_ we want to reflect that in our health check. If we do not stick with the _raised_ severity we would not figure out that the cluster is _red_. I also assume we should aim to be as close to mimic what _Admin-UI_ shows, as it is weird if we report a _YELLOW_ and Admin-UI shows _RED_.

IMHO it would make total sense if crateDB offers an endpoint that can be queried (no sql involved) to report health. How it is implemented now, it seems a bit that this quite a bit delegated to the _front-end_. 

I think we want to stick with the value set by the core-team.

WalBeh:
After testing it on a cluster scaled from '0' -> `1`,  we figured out that (without bouncing pod-0) the first pod still _remembers_ the old expected nodes number and would trigger a `RED` state, where admin-ui is already showing `GREEN`. Which invalidates the approach by selecting `severity` from `sys.nodes_check`.  Also checking in admin ui it seems admin ui does not query `sys.health `at all.

After some disscusion we came up with different approaches like:
- `SELECT max(severity) FROM (
          SELECT MAX(severity) as severity FROM sys.health
         UNION
         select 3 as "severity" from sys.cluster where master_node is null limit 100
     ) sub;`
- `set global indices.recovery.max_bytes_per_sec = '200mb'` but we would need to remember what it was set before the check.

PR #468, opened 2023-02-17
No pre-flight check on PVC expand

WalBeh:
## Summary of changes
Do not perform cluster pre-flights checks when expanding disk.

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changelog
 Unreleased
 ----------
 
+* Do not perform cluster pre-flights checks when expanding disk.
+
 * Fix failing operator tests.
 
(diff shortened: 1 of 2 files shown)


SStorm (approved):
Looks good! We would ideally want a test for it, but we don't currently test for when `NO_DOWNTIME_STORAGE_EXPANSION` is enabled, so I think we can skip it this time.

WalBeh:
> Looks good! We would ideally want a test for it, but we don't currently test for when `NO_DOWNTIME_STORAGE_EXPANSION` is enabled, so I think we can skip it this time.

I tried to test it, with under a _disk_full_ scenario on `dev`.

PR #469, opened 2023-02-20
Disable pre-flight checks for adding replicas

WalBeh:
## Summary of changes
Disable pre-flight checks for adding replicas.

## Checklist

### DIFF
--- a/tests/test_scale.py
+++ b/tests/test_scale.py
@@ -316,14 +316,14 @@ async def test_scale_cluster_while_create_snapshot_running(
 
     await create_fake_cronjob(api_client, name, namespace.metadata.name)
 
-    host, password = await start_cluster(name, namespace, core, coapi, 1)
+    host, password = await start_cluster(name, namespace, core, coapi, 2)
 
     conn_factory = connection_factory(host, password)
     await create_test_sys_jobs_table(conn_factory)
     await insert_test_snapshot_job(conn_factory)
 
     # When
-    await _scale_cluster(coapi, name, namespace, 2)
+    await _scale_cluster(coapi, name, namespace, 1)
 
     # Then
     await assert_wait_for(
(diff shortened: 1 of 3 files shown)


tests/test_scale.py
```diff
@@ -314,49 +311,12 @@ async def test_scale_cluster_while_create_snapshot_running(
     core = CoreV1Api(api_client)
     name = faker.domain_word()
 
-    await create_fake_cronjob(api_client, name, namespace.metadata.name)
```
SStorm, line 317: I would keep these checks, but maybe scale down instead of up, so that it is still interrupted while a snapshot is happening.

tests/test_scale.py
```diff
@@ -314,49 +311,12 @@ async def test_scale_cluster_while_create_snapshot_running(
...
-
     host, password = await start_cluster(name, namespace, core, coapi, 1)
```
SStorm, line 314:
```suggestion
    host, password = await start_cluster(name, namespace, core, coapi, 2)
```

tests/test_scale.py
```diff
@@ -314,49 +311,12 @@ async def test_scale_cluster_while_create_snapshot_running(
...
 
-    conn_factory = connection_factory(host, password)
-    await create_test_sys_jobs_table(conn_factory)
-    await insert_test_snapshot_job(conn_factory)
-
     # When
     await _scale_cluster(coapi, name, namespace, 2)
```
SStorm, line 317:
```suggestion
    await _scale_cluster(coapi, name, namespace, 1)
```

WalBeh:
This needs probably a bit more of investigations, as not all of the testing showed a positive outcome. Scaling a cluster (on the k8s side) without an active master node seem to be a bit problematic.

PR #473, opened 2023-03-01
Rollback CrateDB version on upgrade failures

tomach:
## Summary of changes
- revert `spec.cluster.version` at the end of the execution in case there was an error somewhere in the upgrade process. To avoid kopf getting stuck or ending in an infinite loop setting the value back and forth, the "rollback" is done at the end of the execution, only followed by updating the introduced `spec.cluster.lastUpdatedAt` timestamp. This is recognized by kopf as another change (which is actually ignored afterwards) to the CRD and helps cleaning up the status annotations in the resource.
- move timeout handling from the main `cluster_update` handler to the individual subhandlers. Therefore it is necessary to store the timestamp when each subhandler really started working somewhere (in the `status` object in this case) 

https://github.com/crate/cloud/issues/1079

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@ Unreleased
 
 * Include ``sys.cluster`` for checking cluster healthiness.
 
+* Revert ``spec.cluster.version`` on upgrade failures.
+
 2.22.0 (2023-01-31)
 -------------------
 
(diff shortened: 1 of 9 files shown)


tomach:
closing this one. some parts have been implemented with https://github.com/crate/crate-operator/pull/582

PR #485, opened 2023-03-22
Ignore failed chown on EFS volumes

WalBeh:
## Checklist


crate/operator/create.py
```diff
@@ -572,6 +572,12 @@ def get_statefulset_crate_volume_mounts(
 
 
 def get_statefulset_init_containers(crate_image: str) -> List[V1Container]:
+    heapdump_cmd = (
+        "mkdir -pv /resource/heapdump ; chown -R crate:crate /resource"  # noqa
+    )
+    if config.CLOUD_PROVIDER == CloudProvider.AWS:
+        heapdump_cmd = f"{heapdump_cmd} || true'"
```
SStorm, line 579:
```suggestion
    # Ignore any failures on AWS, where we are likely using EFS (which do not permit chown)
    if config.CLOUD_PROVIDER == CloudProvider.AWS:
        heapdump_cmd = f"{heapdump_cmd} || true'"
```

SStorm (approved):
Just one suggestion for a comment.

Technically it does not imply that AWS == EFS, but we have no way of knowing what the storage class is using in the operator, so I think this is fine.

PR #487, opened 2023-03-22
Add operation type to feedback webhooks payload

tomach:
## Checklist


crate/operator/operations.py
```diff
@@ -621,6 +625,9 @@ async def suspend_or_start_cluster(
                     logger=logger,
                     status=WebhookStatus.IN_PROGRESS,
                     operation=WebhookOperation.UPDATE,
+                    action=WebhookAction.SUSPEND
```
SStorm, line 628: IIRC the action is also `SUSPEND` if either we're suspending or resuming the cluster?..

SStorm (approved):
Just one Q about the suspend operation, but LGTM!

crate/operator/operations.py
tomach: yes, type `SUSPEND` is for both suspending and resuming in the API.

SStorm: I mean, if we're doing a suspend operation, then old_replicas will be 1 and we'll return action=SCALE?..

tomach: no, this is the `if old_replicas < new_replicas:` block (resume). The suspend part is handled in the `elif old_replicas > new_replicas:` block further down.

SStorm: Ah ok! Got it :)

PR #492, opened 2023-04-04
Suspending/resuming a cluster takes down / creates the load balancer

juanpardo:
## Checklist

### DIFF
--- a/crate/operator/create.py
+++ b/crate/operator/create.py
@@ -985,20 +985,27 @@ async def create_services(
 ) -> None:
     async with ApiClient() as api_client:
         core = CoreV1Api(api_client)
-        await create_data_service(
+
+        # Create the load balancer service
+        svc = get_data_service(
             owner_references,
-            namespace,
             name,
             labels,
             http_port,
             postgres_port,
-            transport_port,
             dns_record,
-            logger,
             source_ranges,
             additional_annotations=additional_annotations,
         )
+        await call_kubeapi(
+            core.create_namespaced_service,
+            logger,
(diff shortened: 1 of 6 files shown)


crate/operator/create.py
```diff
@@ -1009,6 +1010,98 @@ async def create_services(
         )
 
 
+async def create_data_service(
```
SStorm, line 1013: Since `create_services()` is idempotent, I'm not sure we need to split this out - you could just call `create_services()` and it will re-create the LoadBalancer and not touch the discovery service.

crate/operator/create.py
```diff
@@ -1009,6 +1010,98 @@ async def create_services(
...
+                logger,
+                continue_on_conflict=True,
+                namespace=namespace,
+                body=svc,
+            )
+        except Exception as e:
+            raise e
+
+
+async def create_lb_service(
```
SStorm, line 1051: Maybe call it `recreate_services()` and  get it to call `create_services()`?

crate/operator/operations.py
```diff
@@ -681,6 +696,73 @@ async def suspend_or_start_cluster(
...
+
+    if svc:
+        svc_name = f"crate-{name}"
+        await core.delete_namespaced_service(
+            name=svc_name, namespace=namespace, body=V1DeleteOptions()
+        )
+
+    # If the Load balancer service was already not present or a delete call did not
+    # trigger an exception we consider the operation a success
+    return True
```
SStorm, line 725: Might as well not returning anything, if it's always True? Nothing is checking this value either.

PR #501, opened 2023-04-16
Upgraded sql_exporter so it will load on arm64 systems

robd003:
## Summary of changes
Upgraded sql_exporter to new version that has arm64 support

## Checklist

- [X] Relevant changes are reflected in `CHANGES.rst`
- [X] Added or changed code is covered by tests
- [X] Documentation has been updated if necessary
- [X] Changed code does not contain any breaking changes (or this is a major version change)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changelog
 Unreleased
 ----------
 
+* Upgraded sql_exporter with arm64 support
+
 2.26.1 (2023-04-12)
 -------------------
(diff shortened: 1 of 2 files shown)


robd003:
@SStorm just a heads up that this was the only blocking issue when deploying on AWS Graviton systems. After updating sql_exporter everything works perfectly!

SStorm (approved):
Hi @robd003, many thanks for the contribution!

PR #539, opened 2023-07-19
Fixed deleting a cluster while restoring/cloning

juanpardo:
## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changelog
 Unreleased
 ----------
 
+* Fixed a bug that lead to the namespace not being deleted after deleting a cluster
+  that had a snapshot restore/clone operation in progress.
+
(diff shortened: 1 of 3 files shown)


SStorm (approved):
Very nice solution! Should we make this generic and perhaps move this check into each of the top level handlers in [main.py](https://github.com/crate/crate-operator/blob/master/crate/operator/main.py#L144)? Even though we've only seen this happen for restore operations, I suspect it can theoretically happen for any?

juanpardo:
> Very nice solution! Should we make this generic and perhaps move this check into each of the top level handlers in [main.py](https://github.com/crate/crate-operator/blob/master/crate/operator/main.py#L144)? Even though we've only seen this happen for restore operations, I suspect it can theoretically happen for any?

I was expecting you to say that :)
Will add it to this PR

PR #540, opened 2023-08-07
Add support for restoring metadata from snapshot

tomach:
## Checklist


deploy/charts/crate-operator-crds/templates/cratedbs-cloud-crate-io.yaml
```diff
@@ -291,12 +291,44 @@ spec:
                       snapshot:
                         description: The name of the snapshot to use.
                         type: string
+                      type:
+                        type: string
+                        enum: ["all", "tables", "metadata", "partitions", "sections"]
```
SStorm, line 296: It's not entirely clear to me from the crate docs how "sections" works. Can you restore sections on their own, or do we have to combine this with restoring metadata, and then specifying which section of metadata to restore (i.e. `udfs`)?

tomach: As far as I understand you can restore one (or more) data sections on their own. e.g. `RESTORE SNAPSHOT repository_name.snapshot_name TABLES,VIEWS,USERS,PRIVILEGES,...`. It also says `A section cannot be combined with the ALL, METADATA, or TABLE keywords.` in the docs.

crate/operator/restore_backup.py
```diff
@@ -74,6 +75,11 @@
 DEFAULT_MAX_BYTES_PER_SEC: str = "40mb"
 DEFAULT_CLUSTER_CONCURRENT_REBALANCE: int = 2
 CRASH_COMMAND_DELAY: int = 30
+RESTORE_TYPE_TABLES: str = "tables"
```
juanpardo, line 78: Shouldn't this be an `Enum`, perhaps?

crate/operator/restore_backup.py
```diff
@@ -359,6 +365,140 @@ async def shards_recovery_in_progress(
...
+
+@RestoreType.register_subclass(RESTORE_TYPE_METADATA)
+class RestoreMetadata(RestoreType):
+    def get_restore_keyword(self):
+        return "METADATA"
+
+    async def validate_restore_complete(
+        self, *, cursor: Cursor, snapshot: str, logger: logging.Logger, **_kwargs
+    ):
+        return True
```
juanpardo, line 449: Should this method really invariably return `True`? Would that still be the case if the operation failed for any reason?

tomach: `validate_restore_complete()` only runs after the actual restore operation is finished, e.g. to check if all tables have been restored successfully for operations which include table-restores. For metadata I do not really know how to validate if the metadata has been restored successfully, so in this case it is considered successful as soon as the check for `sys.snapshot_restore` table [here](https://github.com/crate/crate-operator/blob/master/crate/operator/restore_backup.py#L195) succeeds.

If the restore operation itself fails, it still reports the error already before.

crate/operator/restore_backup.py
tomach: yes, good point. thanks! I've changed it.

PR #544, opened 2023-08-29
Increase timeout for querying snapshots table

tomach:
## Summary of changes
Fixes https://github.com/crate/cloud/issues/1362

## Checklist

### DIFF
--- a/crate/operator/restore_backup.py
+++ b/crate/operator/restore_backup.py
@@ -292,28 +292,31 @@ async def get_source_backup_repository_data(
 
 
 async def get_snapshot_tables(
-    cursor: Cursor, snapshot: str, logger: logging.Logger
+    conn_factory, snapshot: str, logger: logging.Logger
 ) -> List[Any]:
     """
     Returns a list of tables included in a snapshot.
 
-    :param cursor: A database cursor to a current and open database connection.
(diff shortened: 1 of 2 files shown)


crate/operator/restore_backup.py
```diff
@@ -871,16 +879,15 @@ async def handle(  # type: ignore
                 get_host(core, namespace, name),
             )
             conn_factory = connection_factory(host, password)
-            async with conn_factory() as conn:
-                async with conn.cursor() as cursor:
```
SStorm, line 716: Could have also set the timeout globally here perhaps? Would have made a smaller diff! :)

SStorm (approved):

crate/operator/restore_backup.py
tomach: yes, but then the timeout would be increased for everything in `validate_restore_complete()` like querying `sys.shards` etc. not sure if we want that? 

The diff should become smaller once I merge the other branch into master - I just needed that code as a base here :)

SStorm: I don't think it matters that much, the one minute timeout is a bit arbitrary, same as 2 minutes! Up to you entirely, I already approved it.

PR #549, opened 2023-09-06
Maintenance: Add support for Python 3.11

amotl:
## About
Make the operator support Python 3.11.

## Blocked by
- [x] Let's do Python 3.10 first: GH-690
- [x] Please merge that first: GH-654

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
     install_requires=[
         "aiopg==1.4.0",
         "bitmath==1.3.3.1",
-        "kopf==1.35.6",
+        "kopf==1.36.2",
         "kubernetes-asyncio==31.1.0",
         "PyYAML<7.0",
         "prometheus_client==0.21.1",
@@ -80,12 +80,13 @@ setup(
             "mypy==1.13.0",
         ],
     },
-    python_requires=">=3.10",
+    python_requires=">=3.10,<3.12",
     classifiers=[
         "Development Status :: 5 - Production/Stable",
         "License :: OSI Approved :: Apache Software License",
         "Programming Language :: Python :: 3",
         "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3.11",
     ],
     use_scm_version=True,
 )
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -13,13 +13,20 @@ concurrency:
 jobs:
   lint:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -49,17 +49,16 @@ def read(path: str) -> str:
     package_data={"crate.operator": ["data/*"]},
     setup_requires=["setuptools>=58", "setuptools_scm>=6.2"],
     install_requires=[
-        "aiopg==1.4.0",
-        "bitmath==1.3.3.1",
-        "kopf==1.35.6",
+        "aiopg<1.5",
```
SStorm, line 52: Due to there not being a lockfile for setuptools, I am strongly against pinning version ranges - this eventually leads to some library releasing a patch version that breaks backwards compatibility accidentally and you no longer have predictable builds. Let's leave it like it was please.

SStorm (commented):
Can we please go with 3.10 first - 3.11 is still a bit on the bleeding edge with some libraries being incompatible.

amotl:
Hi Romanas. I was just playing around a bit, and I do not plan to submit this patch as a proposal for integration.

SStorm:
I only have anecdotal evidence, i.e. when we were trying to upgrade to 3.11 in the [cloud-api](https://github.com/crate/cloud-api/pull/2484), there were issues with celery. That made me come to a mental conclusion "3.11 is too early for us" and forget about it.

amotl:
References about mypy modernizations.

- https://github.com/python/mypy/issues/10600#issuecomment-859225825
- https://github.com/python/mypy/issues/10768

setup.py
amotl: I've removed the pinning of version ranges again.

goat-ssh:
@amotl do you plan working on this one ?

amotl:
@goat-ssh: I can give it another refresh if there is interest.

amotl:
> References about mypy modernizations [...]

- See https://github.com/crate/crate-operator/pull/654.

plaharanne (commented):
It looks good to me, but I would rather let @tomach approve it, as he worked on the other python upgrades, etc.

amotl:
Thanks. @tomach: 🙏?

tomach:
> Do you have any objections?

@amotl No objections from my side, thanks a lot! just to clarify - are we planning to update the Dockerfile to Python 3.11 in a future PR? Right now, this ensures compatibility in CI but doesn’t impact the runtime environment right? or maybe we should also let the integration tests on k8s run with 3.11 [here](https://github.com/crate/jenkins-dsl/pull/1266/files) as a first step?

tomach (approved):
LGTM!

PR #553, opened 2023-10-25
Add `nodepool` to the CRD

tomach:
## Summary of changes
Add optional `nodepool` to the node's CRD and set node affinity and tolerations accordingly when creating a cluster or changing its compute.
https://github.com/crate/cloud/issues/1502

## Checklist

### DIFF
--- a/crate/operator/create.py
+++ b/crate/operator/create.py
@@ -92,6 +92,7 @@ from crate.operator.constants import (
     SHARED_NODE_TOLERATION_KEY,
     SHARED_NODE_TOLERATION_VALUE,
     CloudProvider,
+    Nodepool,
     Port,
 )
 from crate.operator.utils import crate, quorum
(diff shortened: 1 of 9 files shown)


crate/operator/create.py
```diff
@@ -1124,9 +1122,12 @@ def is_shared_resources_cluster(node_spec: Dict[str, Any]) -> bool:
         cpu_limit = node_spec["resources"].get("limits", {}).get("cpu")
         memory_request = node_spec["resources"].get("requests", {}).get("memory")
         memory_limit = node_spec["resources"].get("limits", {}).get("memory")
+        nodepool = node_spec.get("nodepool")
         if not (cpu_request or memory_request):
             return False
-        return cpu_request != cpu_limit or memory_request != memory_limit
+        return (
```
SStorm, line 1128: I guess the question is whether we need to check the requests/limits at all anymore? It might be a bit counter-intuitive, because if your requests/limits do not differ, but you still specify shared for the nodepool, it will ignore that and put it to the dedicated one?...

SStorm (approved):
If I understand correctly, if nodepool is not specified, we will always default to dedicated, right?

crate/operator/create.py
tomach: hm you're right. I can't think of a case where both checks are needed either.

> If I understand correctly, if nodepool is not specified, we will always default to dedicated, right?

yes. if no nodepool is specified, it defaults to the tolerations and affinity of the dedicated one.

PR #572, opened 2024-01-09
Suspend and resume grand central Deployment when a cluster is suspended/resumed

tomach:
## Summary of changes
- Suspend and resume grand central `Deployment` when a cluster is suspended/resumed
- Use the `system` user in gc to connect to crateDB
- Increased the initial delay to avoid crashes while crateDB is starting up
- Pinned `alabaster` version, because the latest does not work with our version of `Sphinx`

https://github.com/crate/cloud/issues/1619

## Checklist

### DIFF
--- a/crate/operator/grand_central.py
+++ b/crate/operator/grand_central.py
@@ -73,6 +73,7 @@ from crate.operator.constants import (
     SHARED_NODE_TOLERATION_EFFECT,
     SHARED_NODE_TOLERATION_KEY,
     SHARED_NODE_TOLERATION_VALUE,
+    SYSTEM_USERNAME,
 )
 from crate.operator.utils import crate
 from crate.operator.utils.kopf import StateBasedSubHandler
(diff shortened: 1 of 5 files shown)


crate/operator/grand_central.py
```diff
@@ -161,7 +159,7 @@ def get_grand_central_deployment(
                                     path="/api/health",
                                     port=GRAND_CENTRAL_BACKEND_API_PORT,
                                 ),
-                                initial_delay_seconds=60,
+                                initial_delay_seconds=180,
```
SStorm, line 162: Was 60 not enough?

crate/operator/operations.py
```diff
@@ -600,6 +602,7 @@ async def suspend_or_start_cluster(
     old: kopf.Body,
     data_diff_items: kopf.Diff,
     scale_backup_metrics: bool,
+    scale_grand_central: bool,
```
SStorm, line 605: Wonder if we need this param at all? Why scale it conditionally?

SStorm (approved):
:100:

crate/operator/grand_central.py
tomach: Sometimes it takes longer to provision the cluster. So I think it makes sense to wait a bit longer before doing the liveness check..

crate/operator/operations.py
tomach: hm, the idea was to not scale it [here](https://github.com/crate/crate-operator/blob/master/crate/operator/operations.py#L1165) and [here](https://github.com/crate/crate-operator/blob/master/crate/operator/operations.py#L1215) (similar to the backup metrics deployment). But actually it might not be necessary, I'll change it..

PR #579, opened 2024-01-26
Limit system user login

WalBeh:
## Summary of changes
~~Limit system user login to 10.0.0.0/8.~~

I assumed that the configuration snippet from below limits the  system  to login from 10.0.0.0/8 only. But it seems like the 99.method=password  spoils my plan!
```
 82     - -Cauth.host_based.enabled=true
 83     - -Cauth.host_based.config.0.user=crate
 84     - -Cauth.host_based.config.0.address=_local_
 85     - -Cauth.host_based.config.0.method=trust
 86     - -Cauth.host_based.config.1.user=system
 87     - -Cauth.host_based.config.1.address=10.0.0.0/8
 88     - -Cauth.host_based.config.1.method=password
 89     - -Cauth.host_based.config.99.method=password
 ```
 
In can see the real IP of the connection (=external IP) `system` can still login, due to the setting in _line 89_  where _password login from everywhere_ is allowed.

## Checklist

### DIFF
--- a/crate/operator/create.py
+++ b/crate/operator/create.py
@@ -421,7 +421,7 @@ def get_statefulset_crate_command(
         "-Cauth.host_based.config.0.address": "_local_",
         "-Cauth.host_based.config.0.method": "trust",
         "-Cauth.host_based.config.1.user": "system",
-        "-Cauth.host_based.config.1.address": "10.0.0.0/8",
+        "-Cauth.host_based.config.1.address": config.ALLOWED_SYSTEM_USER_CIDR,
         "-Cauth.host_based.config.99.method": "password",
         "-Cpath.data": ",".join(
             f"/data/data{i}" for i in range(node_spec["resources"]["disk"]["count"])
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: crate/operator/create.py [hunk 1]
```diff
@@ -420,6 +420,8 @@ def get_statefulset_crate_command(
         "-Cauth.host_based.config.0.user": "crate",
         "-Cauth.host_based.config.0.address": "_local_",
         "-Cauth.host_based.config.0.method": "trust",
+        "-Cauth.host_based.config.1.user": "system",
+        "-Cauth.host_based.config.1.address": "10.0.0.0/8",
```
SStorm, line 424: We should probably make this configurable, since no guarantee that all k8s is using class A subnets? Also people might want to restrict it further, if they use a different subnet mask in their k8s.

WalBeh: > We should probably make this configurable, since no guarantee that all k8s is using class A subnets? Also people might want to restrict it further, if they use a different subnet mask in their k8s.

_configurable_ as in CRD, plus a UI?

deploy/charts/crate-operator-crds/templates/cratedbs-cloud-crate-io.yaml
```diff
@@ -173,6 +173,9 @@ spec:
                     items:
                       type: string
                     type: array
+                  allowedSystemUserCIDR:
```
SStorm, line 176: You don't need this, if the setting is coming from the environment - we don't need to be able to override it for individual CrateDBs?

PR #582, opened 2024-01-30
add timeout to cluster update handlers

plaharanne:
## Summary of changes

Moved timeout handling from the main `cluster_update` handler to the individual subhandlers. Therefore it is necessary to store the timestamp when each subhandler really started working somewhere (in the status object in this case).
It has a better granularity and allows to have for example a timeout after 24h for the pre-upgrade tasks and I kept 2 hours for the upgrade itself).
Code based on https://github.com/crate/crate-operator/pull/473

https://github.com/crate/cloud/issues/1621

## Checklist

### DIFF
--- a/crate/operator/upgrade.py
+++ b/crate/operator/upgrade.py
@@ -170,6 +170,7 @@ async def upgrade_cluster(
 
 class UpgradeSubHandler(StateBasedSubHandler):
     @crate.on.error(error_handler=crate.send_update_failed_notification)
+    @crate.timeout(timeout=float(config.CLUSTER_UPDATE_TIMEOUT))
     async def handle(  # type: ignore
         self,
         namespace: str,
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ Unreleased
 
 * Added the `Authorization` header to the allowed list for GC.
 
+* Moved cluster update timeout to the handlers level.
+
 2.34.1 (2024-02-06)
 -------------------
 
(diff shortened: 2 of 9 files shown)


juanpardo (approved):

SStorm (approved):
Can you elaborate a bit what was the problem with the timeout on the entire `cluster_update` handler?

plaharanne:
> Can you elaborate a bit what was the problem with the timeout on the entire `cluster_update` handler?

@SStorm It is not really a problem, it is just that the whole `cluster_update` timeout after 2 hours, and therefore the timeout on the handlers level can only be shorter. I could just increase this value, it would work fine.
By moving it the the handler level, it has a better granularity (e.g. I put 24h for the pre-upgrade tasks and I kept 2 hours for the upgrade itself).

crate/operator/upgrade.py
```diff
@@ -170,6 +170,7 @@ async def upgrade_cluster(
 
 class UpgradeSubHandler(StateBasedSubHandler):
     @crate.on.error(error_handler=crate.send_update_failed_notification)
+    @crate.timeout(timeout=float(config.CLUSTER_UPDATE_TIMEOUT))
```
tomach, line 173: maybe I'm wrong, but doesn't the runtime for each subhandler already start as soon as they are registered? So this one might be delayed while it is waiting for the `BeforeClusterUpdateSubHandler` with a temporary error [here](https://github.com/crate/crate-operator/blob/master/crate/operator/utils/kopf.py#L137) but for kopf it seems to be running all the time, so the whole process might still time out after 2h already?

plaharanne: good point, let me test it

tomach (approved):
LGTM! :100:

PR #593, opened 2024-02-08
Delayed cronjob re-enabling after cluster upgrade/resume

plaharanne:
## Summary of changes

The `AfterClusterUpdateSubHandler` is now split in two. The first one is as it used to be and the second one is to re-enable the cronjob.
This new handler is not added to the `depends_on` list because it can be delayed by 1h in some cases. It means that the success notification will be sent and the handler will be executed later (if an issue occur, the `error` decorator will do the trick). 
The delay is applied if `delay_cronjob=True` is in the status. It is set to True in the `upgrade` handler.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/1621

### DIFF
--- a/crate/operator/main.py
+++ b/crate/operator/main.py
@@ -54,6 +54,9 @@ from crate.operator.handlers.handle_update_user_password_secret import (
 from crate.operator.handlers.handle_upgrade_grand_central import upgrade_grand_central
 from crate.operator.kube_auth import login_via_kubernetes_asyncio
 from crate.operator.operations import (
+    DELAY_CRONJOB,
+    DELAY_CRONJOB_START,
+    ensure_cronjob_reenabled,
     is_namespace_terminating,
     update_sql_exporter_configmap,
 )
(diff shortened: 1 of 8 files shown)


crate/operator/operations.py
```diff
@@ -1091,24 +1092,65 @@ async def handle(  # type: ignore
...
+        name: str,
+        body: kopf.Body,
+        old: kopf.Body,
+        patch: kopf.Patch,
+        status: kopf.Status,
+        logger: logging.Logger,
+        **kwargs: Any,
+    ):
+        # Delay cronjob after resuming or upgrading a cluster
+        if DELAY_CRONJOB in status.keys() and status.get(DELAY_CRONJOB) is True:
```
juanpardo, line 1134: You can simplify this line by specifying a default:
```
if status.get(DELAY_CRONJOB, False):
```

juanpardo (commented):
Looks good to me :)

Aren't any tests affected by not immediately re-enabling the snapshot job?

crate/operator/main.py
```diff
@@ -344,3 +347,36 @@ async def resume_sql_exporter_configmap(
     """
     await raise_on_namespace_terminating(namespace)
     await update_sql_exporter_configmap(namespace, name, logger)
+
+
+@kopf.timer(
```
SStorm, line 352: Cool solution :+1:

tomach (approved):
LGTM! :100:

PR #599, opened 2024-02-21
Bootstrap gc_admin user

Taliik:
## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/1677

### DIFF
--- a/crate/operator/create.py
+++ b/crate/operator/create.py
@@ -1104,6 +1104,20 @@ def get_system_user_secret(
     )
 
 
+def get_gc_user_secret(
+    owner_references: Optional[List[V1OwnerReference]], name: str, labels: LabelType
+) -> V1Secret:
+    return V1Secret(
+        data={"password": b64encode(gen_password(50))},
+        metadata=V1ObjectMeta(
+            name=f"user-gc-{name}",
+            labels=labels,
+            owner_references=owner_references,
+        ),
+        type="Opaque",
+    )
+
+
 async def create_system_user(
     owner_references: Optional[List[V1OwnerReference]],
     namespace: str,
--- a/crate/operator/grand_central.py
+++ b/crate/operator/grand_central.py
@@ -65,8 +65,10 @@ from kubernetes_asyncio.client import (
 )
 from kubernetes_asyncio.client.api_client import ApiClient
 
+from crate.operator.bootstrap import bootstrap_gc_admin_user
 from crate.operator.config import config
 from crate.operator.constants import (
+    GC_USERNAME,
     GRAND_CENTRAL_BACKEND_API_PORT,
     GRAND_CENTRAL_PROMETHEUS_PORT,
     GRAND_CENTRAL_RESOURCE_PREFIX,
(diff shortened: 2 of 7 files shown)


crate/operator/grand_central.py
```diff
@@ -390,6 +393,33 @@ async def create_grand_central_backend(
...
+            continue_on_conflict=True,
+            namespace=namespace,
+            body=get_gc_user_secret(owner_references, name, labels),
+        )
+
+        # circular import issue
+        from crate.operator.operations import get_cratedb_resource
+
+        # Before trying to create the gc_admin user, we need to make sure
+        # CrateDB is already started and the system user exists.
```
SStorm, line 408: Why do we care? Won't it fail and retry?

We might want to consider splitting this out into a subhandler, so that it wouldn't block creating the deployment, service etc. Cause right now it will effectively delay creating grand central until the cluster finishes bootstrapping.

crate/operator/grand_central.py
```diff
@@ -390,6 +393,33 @@ async def create_grand_central_backend(
...
+        cratedb = await get_cratedb_resource(namespace, name)
+        pod_name = get_crash_pod_name(cratedb, name)
+        scheme = get_crash_scheme(cratedb)
+        command = "SELECT * FROM sys.users WHERE name = 'system';"
+
+        result = await run_crash_command(namespace, pod_name, scheme, command, logger)
+        if "SELECT 1" in result:
+            logger.info("... success")
```
SStorm, line 416: This log message would be pretty random I think (not clear what it refers to)

crate/operator/grand_central.py
SStorm: In general, all handlers in kopf should be idempotent, and should be able to retry if an error happens half way through (so i.e., if you manage to create the user but then creating the deployment fails, it would retry the entire handler - and you don't want creating the gc_admin user to fail if it already exists).

So I think it might be a good idea to do this creation in a subhandler instead. And possibly also move the deployment/service/ingress creation into one too...

Taliik: Makes sense. I will move the user creation into a subhandler and create a follow up PR for deployment/service/ingress/...

crate/operator/create.py
```diff
@@ -1127,6 +1141,28 @@ async def create_system_user(
         )
 
 
+async def create_gc_admin_user(
```
SStorm, line 1144: This is unused I think?

crate/operator/handlers/handle_create_grand_central.py
```diff
@@ -52,3 +56,11 @@ async def create_grand_central(
             ),
             id="create_grand_central",
         )
+
+        kopf.register(
+            fn=subhandler_partial(
+                create_grand_central_user, namespace, name, cratedb["metadata"], logger
+            ),
+            id="create_grand_central_user",
+            backoff=config.BOOTSTRAP_RETRY_DELAY,
```
SStorm, line 65: I suspect you might want to add a `config.BOOTSTRAP_RETRY_DELAY` to the above subhandler too, because it will always fail initially, because of the missing secret.

SStorm (approved):

PR #604, opened 2024-03-05
Try to fix gc tests

tomach:
## Summary of changes
This fixes the failing tests when starting a cluster with gc enabled is blocked and hence timing out.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/1724

### DIFF
--- a/crate/operator/utils/kubeapi.py
+++ b/crate/operator/utils/kubeapi.py
@@ -19,10 +19,10 @@
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
 
-import asyncio
 import logging
 from typing import Awaitable, Callable, Optional
 
+import kopf
 from kubernetes_asyncio.client import ApiException, CoreV1Api, V1ObjectMeta, V1Secret
 
 from crate.operator.config import config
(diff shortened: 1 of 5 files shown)


crate/operator/utils/kubeapi.py
```diff
@@ -191,18 +192,13 @@ async def get_host(core: CoreV1Api, namespace: str, name: str) -> str:
...
+            None,
         )
-
-        if not await is_lb_service_present(core, namespace, name):
-            cratedb = await get_cratedb_resource(namespace, name)
-            await recreate_services(
-                namespace, name, cratedb["spec"], cratedb["metadata"], logger
-            )
+        if not service:
+            raise kopf.TemporaryError("Waiting for service to be created...", delay=5)
```
SStorm, line 201: Maybe better place would be to throw this exception in `get_service_public_hostname`, and  get rid of the for loop there?

tomach: Yes I thought so too, unfortunately `get_service_public_hostname` is also used outside `kopf` [here](https://github.com/crate/crate-operator/blob/master/tests/utils.py#L181) which makes all tests blow up when I raise a `kopf.TemporaryError` in there :thinking:

SStorm: You could make a for loop there instead? :stuck_out_tongue:

SStorm (approved):
:100:

PR #632, opened 2024-07-23
fixed Grand Central initContainer image update

plaharanne:
## Checklist

- [ ] Link to issue this PR refers to: https://github.com/crate/cloud/issues/1974


crate/operator/grand_central.py
```diff
@@ -479,6 +480,13 @@ async def update_grand_central_deployment_image(
         )
         api_container.image = image
```
tomach, line 482:
```suggestion
if deployment.spec.template.spec.init_containers:
```
I was playing around with it too and had an old version of gc deployed (from before the `initContainer` was introduced) and was wondering why it exploded :)

tomach (approved):
LGTM! :100:

PR #654, opened 2024-09-20
Maintenance: Update to mypy 1.11.2

amotl:
## Problem
Updating to Python 3.11 makes mypy croak.
```
fatal error: code.h: No such file or directory
```
-- https://github.com/crate/crate-operator/actions/runs/10957798573/job/30426680335?pr=549#step:6:91

## References
- https://github.com/crate/crate-operator/pull/549


amotl:
Hi. This patch needs a few adjustments to satisfy the type checker. Resolving them probably needs better insights into the code base than what I can provide. /cc @goat-ssh 

https://github.com/crate/crate-operator/actions/runs/10958103493/job/30427634720?pr=654

crate/operator/prometheus.py
```diff
@@ -19,6 +19,8 @@
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
 
+# mypy: disable-error-code="arg-type, attr-defined, list-item, operator"
+
```
amotl, line 23: That is blatantly ignoring a few spots of admonitions by mypy within this file, and should certainly be revisited.

tests/test_metrics.py
```diff
@@ -19,6 +19,8 @@
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
 
+# mypy: disable-error-code="attr-defined, arg-type, union-attr"
+
```
amotl, line 23: That is blatantly ignoring a few spots of admonitions by mypy within this file, and should certainly be revisited.

amotl (commented):
Hi. I've just updated this patch, and CI signals success, so I am handing it in for review. On a few spots it is obvious it needs more love. On others, I am not sure the right adjustment has been applied. On others again, I believe the updates are just right.

In other words, you need to diligently review and eventually take over this patch, to make it converge well. It aimed to make a start on this topic at least, in order to apprach upgrading to support more recent versions of Python.

amotl:
Hi there. Resolving this would unlock a whole bunch of patches stacked upon this one. However, this particular patch needs a few dedicated cycles by authors more accustomed with the code base.

Maybe you can afford to spend a few minutes on this, @tomach? 🙏

tomach:
Hi @amotl! Thanks for the reminder. I’ll be taking a look at this today. Thanks for your effort and patience! 🙏

crate/operator/prometheus.py
amotl: 11b60d0c10be made excellent progress here, thank you so much!

tests/test_metrics.py
amotl: Dito.

tomach:
I tried to fix all mypy errors in `crate/operator/prometheus.py` and `tests/test_metrics.py`. And changed one fix in `crate/operator/handlers/handle_create_cratedb.py` where I think raising an error if `status.get(CLUSTER_CREATE_ID)` is undefined could cause issues, because it might be the case in the first run of a create-operation (but not 100% sure about that).

There is still one issue with sphinx now... :see_no_evil:

amotl:
Thank you very much. I am sure you can also tame the Sphinx, otherwise please ping me about it.

amotl:
You made it, thanks! 💯
Unfortunately, I can't approve the PR because I created it, so we need to ask @juanpardo, @Taliik, or @plaharanne.

tomach:
Yes, thank you! I [added it](https://github.com/crate/crate-operator/pull/654/commits/737a1b8d1d95c0fb5868714f5f188080c79110ad) to the `nitpick_ignore` list. Since it is just an interface and there is no official docs for `prometheus_client` on `readthedocs`, this should be fine I think.

amotl:
I think it's fine. The same thing happens in a [future patch](https://github.com/crate/crate-operator/pull/712/files#diff-008dcb3426febd767787b1521f1fe33086313b927ea37eaab86df5fa88a51698).

plaharanne (approved):
:+1:

PR #675, opened 2024-11-06
Maintenance: Update to Python 3.10

amotl:
## About
Just a humble attempt to modernize the Python version.

## Notes
The patch depends on this one, which needs a bit of more curation by code owners.
- GH-654

PR #690, opened 2024-12-13
Update to Python 3.10

tomach:
## Summary of changes
Thanks a lot for your work @amotl! 
I cherry-picked your commit from GH-675 and updated mypy to a newer version - not the latest one though. I thought a gradual approach might be better to keep things smoother and tackle changes incrementally without adding too much all at once.
So I would leave GH-654 open for now and address it once we tackle the upgrade to Python 3.11. wdyt?

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/2314

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -19,7 +19,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.8
+          python-version: "3.10"
       - name: Install dependencies
         run: |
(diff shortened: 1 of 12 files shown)


plaharanne (approved):
It looks good to me but I don't mind if @amotl gives his feedback too.

amotl (approved):
Every update helps. Thanks. In general, the gradual approach to keep things smoother needs some more velocity on many ends.

PR #717, opened 2025-02-18
Enable global jwt settings in CrateDB 5.10.1+

tomach:
## Summary of changes
This ensures that default global JWT authentication settings are added to the CrateDB command when:

- Deploying new clusters with version >= `5.10.1`
- Upgrading existing clusters to CrateDB version >= `5.10.1`

https://cratedb.com/docs/crate/reference/en/master/config/node.html#jwt-defaults

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/2244

### DIFF
--- a/crate/operator/create.py
+++ b/crate/operator/create.py
@@ -423,6 +423,7 @@ def get_statefulset_crate_command(
     is_master: bool,
     is_data: bool,
     crate_version: str,
+    cloud_settings: Optional[Dict[str, str]],
 ) -> List[str]:
     expected_nodes_setting_name = "gateway.expected_nodes"
     recover_after_nodes_setting_name = "gateway.recover_after_nodes"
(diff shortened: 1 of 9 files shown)


crate/operator/create.py
```diff
@@ -423,6 +423,7 @@ def get_statefulset_crate_command(
     is_master: bool,
     is_data: bool,
     crate_version: str,
+    grand_central_settings: Optional[Dict[str, str]],
```
proddata, line 426:
```suggestion
    cloud_settings: Optional[Dict[str, str]],
```
In theory this not necessarily a grand central setting, but rather cloud setup specifc. Considering that also non cloud users have access to the operator, should this rather be called `cloud_settings`, `cratedb_cloud_settings` or `managed_service_settings`?

tomach: `cloud_settings` seems fine. the `grandCentral` section is only set when deployed via cloud, so we could put all kind of other cloud settings in there as well in the future. I'm going to rename it.

plaharanne (approved):
Looks good to me

proddata (approved):

PR #734, opened 2025-04-28
Handle transient errors gracefully in health check

tomach:
## Summary of changes
This improves the `ping_cratedb_status` handler by catching transient errors (like ClientConnectorError and TimeoutError) and logs them as warnings instead of raising unhandled exceptions.

I also had to upgrade black to support match-case syntax.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/2566

### DIFF
--- a/crate/operator/handlers/handle_ping_cratedb_status.py
+++ b/crate/operator/handlers/handle_ping_cratedb_status.py
@@ -20,8 +20,10 @@
 # software solely pursuant to the terms of the relevant commercial agreement.
 
 import logging
+from asyncio import TimeoutError
 
 import kopf
+from aiohttp.client_exceptions import ClientConnectorError
 from kubernetes_asyncio.client import CoreV1Api
 
 from crate.operator.cratedb import connection_factory, get_healthiness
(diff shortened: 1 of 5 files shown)


crate/operator/handlers/handle_ping_cratedb_status.py
```diff
@@ -57,23 +59,39 @@ async def ping_cratedb_status(
...
+            async with cursor_cm as cursor:
                 healthiness = await get_healthiness(cursor)
                 # If there are no tables in the cluster, get_healthiness returns
                 # none: default to `Green`, as cluster is reachable
                 status = HEALTHINESS_TO_STATUS.get(
                     healthiness, PrometheusClusterStatus.GREEN
                 )
     except Exception as e:
-        logger.warning("Failed to ping cluster.", exc_info=e)
+        match e:
```
plaharanne, line 80: do you know what is the most efficient between using `match` and many `except` block?

plaharanne: or maybe you can build the custom message in the cases and after the `match`, have `logger.warning(f"{error_msg}", e)` just once.

tomach: hm, you're right. match-case is probably not the most efficient solution here. I changed it to a plain except with isinstance checks to keep it simple. Still keeping the unexpected errors separate because in the case of an unexpected error, we still want to log the whole stacktrace

plaharanne (approved):
It looks good to me

widmogrod (approved):
LGTM.
- Checkout the code, test are running on mack
- Code coverage looks good with this functionality

I have a nitpick.
- This PR is shipping two changes - one is "black to support match-case syntax." and second is handling errors gracefully - I would suggest to split them in two PR or two commits. First appraoch allow to have more granular changes and simpler code review, since linter changes are deterministic, and they don't need to block more nuanced changes.

tomach:
> * This PR is shipping two changes - one is "black to support match-case syntax." and second is handling errors gracefully - I would suggest to split them in two PR or two commits. First appraoch allow to have more granular changes and simpler code review, since linter changes are deterministic, and they don't need to block more nuanced changes.

I created https://github.com/crate/crate-operator/pull/735 for the upgrade of `black`

PR #738, opened 2025-05-13
Downgrade Python to 3.12 due to SSL issues

tomach:
## Summary of changes
This PR downgrades the project’s Python version from 3.13 to 3.12 to work around SSL certificate verification issues with AKS k8s API introduced in Python 3.13. The issue appears to stem from stricter behavior in urllib3, which is pulled in indirectly by dependencies

```
SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Missing Authority Key Identifier
```

## Checklist

- [ ] Link to issue this PR refers to:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,7 +90,6 @@ setup(
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
         "Programming Language :: Python :: 3.12",
-        "Programming Language :: Python :: 3.13",
     ],
     use_scm_version=True,
 )
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,7 +2,7 @@
 build:
   os: "ubuntu-22.04"
   tools:
-    python: "3.13"
+    python: "3.12"
 
 python:
   install:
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -90,7 +90,6 @@ def read(path: str) -> str:
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
         "Programming Language :: Python :: 3.12",
-        "Programming Language :: Python :: 3.13",
```
plaharanne, line 93: this line was added by this PR https://github.com/crate/crate-operator/commit/21275f8336d6a5731c258289f6cb539c2a608a3c
I don't know if it's necessary to remove it.

tomach:
> In the previous PR, the version was also changed in .readthedocs.yml, line 5. Otherwise, it looks good to me.

thanks, I also updated it in `.readthedocs.yml`

setup.py
tomach: not sure, but I think it's fine to remove it for now since 3.13 is not (yet) supported.

PR #741, opened 2025-05-16
Set cluster allocation settings during rolling restarts

tomach:
## Summary of changes
- Temporarily set `cluster.routing.allocation.enable` to `new_primaries` before deleting each pod and reset it after the pod has restarted. This ensures more controlled shard allocation during rolling restarts.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/2564

### DIFF
--- a/crate/operator/operations.py
+++ b/crate/operator/operations.py
@@ -516,6 +516,22 @@ async def restart_cluster(
             operation=WebhookOperation.UPDATE,
             action=action,
         )
+        try:
+            conn_factory = await _get_connection_factory(core, namespace, name)
+
+            await set_cluster_setting(
+                conn_factory,
+                logger,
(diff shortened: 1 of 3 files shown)


crate/operator/operations.py
```diff
@@ -557,6 +573,19 @@ async def restart_cluster(
                 patch.status["pendingPods"] = None  # Remove attribute from `.status`
                 return
         else:
+            try:
+                conn_factory = await _get_connection_factory(core, namespace, name)
+
+                await reset_cluster_setting(
```
plaharanne, line 579: it means it will be reset at every retry until the cluster is healthy. Is that ok?

plaharanne (approved):
:+1:

PR #746, opened 2025-06-05
Increase timeouts for rolling restarts, scaling, and updates to 4h

tomach:
## Summary of changes
With the introduction of CrateDB node decommissioning, node shutdown can take significantly longer. To prevent premature failure of rolling restarts or scaling operations, the following operator timeouts have been increased to 4 hours:

- `ROLLING_RESTART_TIMEOUT`
- `SCALING_TIMEOUT`
- `CLUSTER_UPDATE_TIMEOUT`

This ensures sufficient time for cluster operations that involve decommissioning nodes before shutdown, especially under higher load or larger cluster sizes.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/2586

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,9 @@ Unreleased
 
 * Added support to restore a snapshot from a Azure storage.
 
+* Increased rolling restart, scaling, and cluster update timeouts to 4 hours to support
+  longer node shutdown durations introduced by CrateDB decommissioning logic.
+
 2.47.1 (2025-05-14)
 -------------------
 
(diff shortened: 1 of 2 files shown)


tomach:
> Did you search if it is possible to have a timeout based on the number of nodes? If it's not possible, it looks good to me.

I think it's not possible in our current setup, since decorators are evaluated at definition time, long before the handler actually runs and the body is available (which we would need to get the number of nodes).. otherwise we would probably have to implement the whole timeout logic of kopf ourselves, which I'd rather avoid for now :sweat_smile:

PR #837, opened 2026-05-13
Replace grand-central Ingress with HTTPRoute and Traefik Middlewares

tomach:
## Summary of changes
Extends the existing `exposure` field support to grand-central. When `spec.cluster.exposure: traefik`, grand-central is now exposed through the Gateway API (`HTTPRoute`) and three Traefik Middlewares instead of an nginx `Ingress`. The default `loadbalancer` path is unchanged.

- `grand-central.py` adds builders for `HTTPRoute`, `compress-js`, `buffering`, and `ip-allowlist` Middlewares; adds `create_grand_central_exposure` (routing resources only, no deployment/service) and delete helpers for both paths
- `exposure.py` - `ChangeExposureSubHandler` now also switches grand-central resources when the exposure field changes
- `operations.py` - `suspend_or_start_grand_central` deletes routing resources on suspend and recreates them on resume, respecting the active exposure mode
- `handle_update_allowed_cidrs.py` patches the `ip-allowlist` Middleware instead of the Ingress annotation when `exposure=traefik`
- RBAC - adds permissions for `gateway.networking.k8s.io/httproutes` and `traefik.io/middlewares`

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/2905

### DIFF
--- a/crate/operator/exposure.py
+++ b/crate/operator/exposure.py
@@ -36,6 +36,12 @@ from crate.operator.create import (
     _lb_annotations_to_remove,
     get_owner_references,
 )
+from crate.operator.grand_central import (
+    create_grand_central_exposure,
+    delete_grand_central_ingress,
+    delete_grand_central_traefik_resources,
+    read_grand_central_deployment,
+)
 from crate.operator.utils import crate
 from crate.operator.utils.k8s_api_client import GlobalApiClient
 from crate.operator.utils.kopf import StateBasedSubHandler
@@ -614,3 +620,20 @@ class ChangeExposureSubHandler(StateBasedSubHandler):
                 postgres_port,
                 logger,
             )
+
+        # grand-central: only if GC is deployed in this cluster
+        gc_deployment = await read_grand_central_deployment(namespace, name)
+        if gc_deployment:
+            await create_grand_central_exposure(
+                namespace=namespace,
+                name=name,
+                spec=spec,
+                meta=body["metadata"],
+                logger=logger,
+                use_traefik=(new_exposure == "traefik"),
+            )
+
+            if old_exposure == "traefik":
+                await delete_grand_central_traefik_resources(namespace, name, logger)
+            else:
+                await delete_grand_central_ingress(namespace, name, logger)
--- a/crate/operator/grand_central.py
+++ b/crate/operator/grand_central.py
@@ -24,8 +24,10 @@ from typing import Any, Dict, List, Optional
 
 import kopf
 from kubernetes_asyncio.client import (
+    ApiException,
     AppsV1Api,
     CoreV1Api,
+    CustomObjectsApi,
     NetworkingV1Api,
     V1Affinity,
     V1Container,
@@ -89,6 +91,25 @@ from crate.operator.utils.k8s_api_client import GlobalApiClient
 from crate.operator.utils.kubeapi import call_kubeapi
 from crate.operator.utils.secrets import get_image_pull_secrets
 from crate.operator.utils.typing import LabelType
+
+# Gateway API / Traefik CRD constants
+_GATEWAY_API_GROUP = "gateway.networking.k8s.io"
+_GATEWAY_API_VERSION = "v1"
+_HTTPROUTE_PLURAL = "httproutes"
+
+_GC_GATEWAY_NAME: str = "traefik"
+_GC_GATEWAY_NAMESPACE: str = "traefik"
+_GC_GATEWAY_SECTION_NAME: str = "websecure"
+
+_TRAEFIK_GROUP = "traefik.io"
+_TRAEFIK_VERSION = "v1alpha1"
+_MIDDLEWARE_PLURAL = "middlewares"
+
+_MIDDLEWARE_COMPRESS_JS = "grand-central-compress-js"
+_MIDDLEWARE_BUFFERING = "grand-central-buffering"
+_MIDDLEWARE_IP_ALLOWLIST = "grand-central-ip-allowlist"
+_MIDDLEWARE_CORS = "grand-central-cors"
+_OPEN_CIDR = ["0.0.0.0/0", "::/0"]
 
 
 def get_grand_central_labels(name: str, meta: kopf.Meta) -> Dict[str, Any]:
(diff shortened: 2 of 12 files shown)


goat-ssh:
Caught some errors on dev on /auth and /health endpoints:
```
The 'Access-Control-Allow-Origin' header contains multiple values 
'https://console.cratedb-dev.cloud,http://localhost:8000', but only one is allowed.
```
According to the W3C and MDN web specs, the Access-Control-Allow-Origin header can only contain a single origin, the wildcard *, or null. It cannot accept a comma-separated list of multiple origins. When you pass a list, browsers reject it as an invalid value, causing the CORS block.

crate/operator/exposure.py
```diff
@@ -614,3 +620,20 @@ async def handle(
...
+
+        # grand-central: only if GC is deployed in this cluster
+        gc_deployment = await read_grand_central_deployment(namespace, name)
+        if gc_deployment:
+            if old_exposure == "traefik":
+                await delete_grand_central_traefik_resources(namespace, name, logger)
+            else:
+                await delete_grand_central_ingress(namespace, name, logger)
+
+            await create_grand_central_exposure(
```
juanpardo, line 627: If for some reason this fails, GC would become inaccessible. Would it be possible to perform the `delete` after the `create`? This way, even if something was to fail, at least the system would not be that compromised from a user perspective

crate/operator/grand_central.py
```diff
@@ -90,6 +92,25 @@
 from crate.operator.utils.secrets import get_image_pull_secrets
 from crate.operator.utils.typing import LabelType
 
+# Gateway API / Traefik CRD constants
+_GATEWAY_API_GROUP = "gateway.networking.k8s.io"
+_GATEWAY_API_VERSION = "v1"
+_HTTPROUTE_PLURAL = "httproutes"
+
+_GC_GATEWAY_NAME: str = "traefik"
+_GC_GATEWAY_NAMESPACE: str = "traefik"
```
juanpardo, line 101: I think @WalBeh was challenging using this namespace in the meeting. I also think it would be nice if we could use the project namespace (but not sure if it's possible) CC @goat-ssh

crate/operator/grand_central.py
```diff
@@ -270,6 +291,514 @@ def get_grand_central_service(
...
+
+    :param owner_references: Owner references to set on the resource.
+    :param name: The CrateDB custom resource name defining the CrateDB cluster.
+    :param labels: Kubernetes labels to apply to the resource.
+    """
+    body = _build_middleware_base(name, _MIDDLEWARE_BUFFERING, labels, owner_references)
+    body["spec"] = {
+        "buffering": {
+            "maxRequestBodyBytes": 1_073_741_824,
+            "memRequestBodyBytes": 2_097_152,
```
juanpardo, line 511: Where does this 2MB come from? 2MB per request seems very big to me but perhaps it's needed

crate/operator/grand_central.py
```diff
@@ -270,6 +291,514 @@ def get_grand_central_service(
...
+        try:
+            await custom.delete_namespaced_custom_object(
+                group=_GATEWAY_API_GROUP,
+                version=_GATEWAY_API_VERSION,
+                namespace=namespace,
+                plural=_HTTPROUTE_PLURAL,
+                name=f"{GRAND_CENTRAL_RESOURCE_PREFIX}-{name}",
+            )
+        except ApiException as e:
+            if e.status != 404:
```
juanpardo: This is great because it would allow idempotency but I wonder if we should write a warning in the logs

crate/operator/grand_central.py
```diff
@@ -307,45 +836,51 @@ def get_grand_central_ingress(
...
+        "nginx.ingress.kubernetes.io/cors-allow-origin": allow_origin,
+        "nginx.ingress.kubernetes.io/cors-allow-methods": (
+            "GET,POST,PUT,PATCH,OPTIONS,DELETE"
+        ),
+        "nginx.ingress.kubernetes.io/cors-allow-headers": (
+            "Content-Type,Authorization"
+        ),
+        "nginx.ingress.kubernetes.io/cors-max-age": "7200",
+    }
+    if cidrs:  # ← add
```
juanpardo: That comment is a bit random :P

crate/operator/grand_central.py
```diff
@@ -431,15 +968,51 @@ async def create_grand_central_backend(
...
-            continue_on_conflict=True,
-            namespace=namespace,
-            body=get_grand_central_ingress(
-                owner_references, name, labels, hostname, spec
-            ),
-        )
+
+        if use_traefik:
+            custom = CustomObjectsApi(api_client)
+            for mw_body in (
```
juanpardo: I think this code is repeated in `create_grand_central_exposure` at line 756

crate/operator/operations.py
```diff
@@ -850,16 +877,69 @@ async def suspend_or_start_cluster(
...
+    namespace: str,
+    name: str,
+    suspend: bool,
+    logger: logging.Logger,
 ):
+    """
+    Scale the grand-central Deployment to 0 (suspend) or 1 (start) and
+    manage its routing resources accordingly.
+
+    On suspend, the Deployment is scaled to 0 and the active routing resource
```
juanpardo, line 890: This would still not delete the DNS entry, right? I hope not because that way we can avoid the DNS propagation time when resuming. CC @goat-ssh

juanpardo (commented):
Added a few comments. It looks good

crate/operator/operations.py
goat-ssh: | Record                                      | Resource                          | How DNS is created                                                      |
|---------------------------------------------|-----------------------------------|-------------------------------------------------------------------------|
| `*.aks1.eastus.azure.cratedb-dev.net` → **CNAME** | IngressRouteTCP (CrateDB cluster) | CNAME to the regional Traefik LB hostname       |
| `*.gc.aks1.eastus.azure.cratedb-dev.net` → **A**   | HTTPRoute (grand-central)         | external-dns creates an A record directly from the Gateway's resolved IP |

> **Note:** Both records resolve to the same Traefik load balancer IP: **51.8.42.241**

so, yes the tenant grand central DNS is deleted on suspend.

crate/operator/grand_central.py
tomach: :smile: thanks! fixed in https://github.com/crate/crate-operator/pull/837/changes/6a99c3488fd24fc552855f9fe0f2a951bad4993a

crate/operator/grand_central.py
tomach: `memRequestBodyBytes` is the in-memory buffering threshold before Traefik spills to disk, not the total allowed request size. the 2 MiB value is Traefik's default afaik.

crate/operator/exposure.py
tomach: Good point! fixed in https://github.com/crate/crate-operator/pull/837/changes/ab7f3a0d5d6051e79a77777d346ac4deecf07ef7

crate/operator/grand_central.py
tomach: Used `logger.info` rather than `logger.warning` because a `404` during cleanup of idempotent resources is expected I think - https://github.com/crate/crate-operator/pull/837/changes/8b722cea10010d2c9842ac629be9fc8b3b1870d7

crate/operator/grand_central.py
tomach: fixed in https://github.com/crate/crate-operator/pull/837/changes/6f7be821ee2eb187070d29bcf87a38bb4cac88e7

crate/operator/operations.py
tomach: I think it's better to keep the current behaviour (delete routing resources on suspend) for consistency with how the nginx Ingress is handled on suspend already - both are deleted and recreated on resume. In practice our clusters are up and healthy in under 2 minutes on resume, so the DNS wait is not a significant concern.

juanpardo (approved):

goat-ssh:
@tomach another angle about proper ordering of the Traefik Middlewares:
```yaml
filters:
  # 1. GUARD: Drop unauthorized traffic immediately to save resources
  - extensionRef:
      group: traefik.io
      kind: Middleware
      name: grand-central-ip-allowlist
    type: ExtensionRef

  # 2. SANITIZE/TRANSFORM: Apply security headers and buffering
  - responseHeaderModifier:
      set:
      - name: X-Frame-Options
        value: DENY
      # ... (rest of your headers)
    type: ResponseHeaderModifier

  - extensionRef:
      group: traefik.io
      kind: Middleware
      name: grand-central-cors
    type: ExtensionRef

  - extensionRef:
      group: traefik.io
      kind: Middleware
      name: grand-central-buffering
    type: ExtensionRef

  # 3. OPTIMIZE: Compression should be last to act on the finalized response payload
  - extensionRef:
      group: traefik.io
      kind: Middleware
      name: grand-central-compress-js
    type: ExtensionRef
```

tomach:
> @tomach another angle about proper ordering of the Traefik Middlewares:
> 
> ```yaml
> filters:
>   # 1. GUARD: Drop unauthorized traffic immediately to save resources
>   - extensionRef:
>       group: traefik.io
>       kind: Middleware
>       name: grand-central-ip-allowlist
>     type: ExtensionRef
> 
>   # 2. SANITIZE/TRANSFORM: Apply security headers and buffering
>   - responseHeaderModifier:
>       set:
>       - name: X-Frame-Options
>         value: DENY
>       # ... (rest of your headers)
>     type: ResponseHeaderModifier
> 
>   - extensionRef:
>       group: traefik.io
>       kind: Middleware
>       name: grand-central-cors
>     type: ExtensionRef
> 
>   - extensionRef:
>       group: traefik.io
>       kind: Middleware
>       name: grand-central-buffering
>     type: ExtensionRef
> 
>   # 3. OPTIMIZE: Compression should be last to act on the finalized response payload
>   - extensionRef:
>       group: traefik.io
>       kind: Middleware
>       name: grand-central-compress-js
>     type: ExtensionRef
> ```

@goat-ssh Good point, reordered the filters so the IP allowlist runs first as a guard and compression runs last - https://github.com/crate/crate-operator/pull/837/changes/fc61577d1002245f25e7d2371c481931bfd34fb0

goat-ssh (approved):

goat-ssh:
@tomach all looking good on dev! let's ship it

PR #846, opened 2026-06-18
Fix CI flakes — cancel leaked ping timers via clean test teardown

WalBeh:
**WIP / draft** — needs a nightly to validate under real parallel AKS load.

## Problem
Nightly k8s runs fail randomly on a *different*, unrelated test each run (grand-central HTTPRoute, exposure bootstrap, restore, change-compute) — none are master-node tests. Ramping AKS worker CPU/RAM and going 3→6 nodes did **not** help, because both causes are **control-plane** bottlenecks, not worker compute.

## Root cause (verified across nightly logs #2466/#2467/#2471)
The session-scoped, all-namespace operator registers a per-cluster `@kopf.timer` that pings every 5s in tests. The `namespace` fixture tore down **fire-and-forget** — deleting the namespace without deleting the CrateDB CR through kopf — so the finalizer was never cleared and the **ping timer was never cancelled**. It then kept hitting the API server for a phantom cluster (`get_host` → 404 → webhook → 404 event post) for the rest of the run. The 404s carry Kubernetes Priority & Fairness headers: the API server throttles the operator → kopf's watch lags → it misses more deletes → more timers leak.

Zombie pings scale super-linearly with run length and track the failure count:

| Run | Duration | Result | Zombie health-checks | Phantom-CR 404s |
|-----|---------:|--------|---------------------:|----------------:|
| #2471 | 17:02 | 1 failed / 287 | 220 | 312 |
| #2466 | 24:22 | 2 failed / 323 | 687 | 1008 |
| #2467 | 27:26 | 2 failed / 328 | **1761** | **2624** |

They starve the process-wide `GlobalApiClient` `Semaphore(10)` that real handlers under test depend on, so whichever test needs the operator at peak contention loses the race.

## Fix (this PR — mechanism 1)
- `namespace` fixture now deletes the CrateDB CR(s) **through kopf** and waits for finalization (best-effort: logs instead of failing a green test) before deleting the namespace, so the ping timer is cancelled.
- `ping_cratedb_status` bails quietly on a 404 (resource/namespace gone): no status report, no webhook, no event-post spam — defense-in-depth for any timer that briefly outlives its cluster.

No master-node logic touched; this is a master-wide harness fix.

## Still TODO (mechanism 2 — separate, needs infra)
A second mechanism shows up as `Pending: unbound immediate PersistentVolumeClaims` (azurefile **Immediate** binding can't keep up with parallel StatefulSet creation, 2 PVCs/pod). Fix is a `WaitForFirstConsumer` StorageClass on the AKS test cluster — can't be validated on minikube, so deferred to a follow-up + nightly.

### DIFF
--- a/crate/operator/handlers/handle_ping_cratedb_status.py
+++ b/crate/operator/handlers/handle_ping_cratedb_status.py
@@ -82,7 +82,7 @@ async def ping_cratedb_status(
             # been deleted but kopf's watch has not yet stopped this timer. There
             # is nothing to report: skip the status update and the webhook so we
             # don't spam the API server with health checks and 404 event posts
-            # for a phantom cluster (see CI_INVESTIGATION.md, mechanism 1).
+            # for a phantom cluster.
             logger.debug("CrateDB resource is gone; skipping health check.")
             return
         logger.warning(
--- a/crate/operator/handlers/handle_ping_cratedb_status.py
+++ b/crate/operator/handlers/handle_ping_cratedb_status.py
@@ -25,6 +25,7 @@ from asyncio import TimeoutError
 import kopf
 from aiohttp.client_exceptions import ClientConnectorError
 from kubernetes_asyncio.client import CoreV1Api
+from kubernetes_asyncio.client.exceptions import ApiException
 
 from crate.operator.cratedb import connection_factory, get_healthiness
 from crate.operator.prometheus import PrometheusClusterStatus, report_cluster_status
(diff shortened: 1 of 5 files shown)


crate/operator/handlers/handle_ping_cratedb_status.py
```diff
@@ -75,6 +76,19 @@ async def ping_cratedb_status(
                 status = HEALTHINESS_TO_STATUS.get(
                     healthiness, PrometheusClusterStatus.GREEN
                 )
+    except ApiException as e:
+        if e.status == 404:
+            # The CrateDB resource (or its namespace) is gone -- the cluster has
+            # been deleted but kopf's watch has not yet stopped this timer. There
+            # is nothing to report: skip the status update and the webhook so we
+            # don't spam the API server with health checks and 404 event posts
+            # for a phantom cluster (see CI_INVESTIGATION.md, mechanism 1).
```
tomach: maybe clean up the comments and remove the references to `CI_INVESTIGATION.md` everywhere, since this seems to be a local file?

tomach (approved):
Cool stuff! 💯

crate/operator/handlers/handle_ping_cratedb_status.py
WalBeh: Good catch — removed the `CI_INVESTIGATION.md` references from all code comments in c2fa738. The inline explanations stay.

PR #847, opened 2026-06-19
Stabilise e2e/k8s tests: bootstrap readiness + retry-on-failure

WalBeh:
Independent CI/robustness fixes extracted ahead of the dedicated-master-nodes feature, so master gets them without waiting for that PR. All three help master's existing setup; none contain master-node logic.

## 1. Wait for the crate container to be ready before the bootstrap exec
The system-user bootstrap execs into the `crate` container (`CREATE USER "system"`). When it fires before the container is ready, the apiserver rejects the exec upgrade with a `400/500 "Invalid response status"`; the handler then retries on kopf's cadence, adding minutes to cluster bring-up and intermittently cascading into timeouts in unrelated, timing-sensitive tests.

Gate the exec on readiness: poll the crate container's ready status first and only exec once it's up. Generous per-attempt budget (300s) since readiness routinely takes a while under load; it's only a per-attempt cap (kopf still retries up to `BOOTSTRAP_TIMEOUT`). On timeout the pod's phase + per-container/init-container state is logged so a stuck start is diagnosable.

(Includes the bootstrap pod-exec failure diagnostics it builds on.)

## 2. Retry k8s/e2e tests on failure
The k8s suite spins real clusters on shared cloud infra in parallel; cluster operations have variable latency (disk attach/resize, apiserver Priority & Fairness throttling), so a fixed-timeout e2e test occasionally loses the race even when the operator is correct.

Add `pytest-rerunfailures` and attach a `flaky(reruns=2)` marker to every `k8s`-marked test when running against a cluster. A transient flake passes on a rerun and keeps the run green; a real bug fails all 3 attempts and still reds the build. Unit tests are never retried, so logic regressions fail immediately. Marker-based — no Jenkins CLI change; GitHub Actions is unaffected (it skips k8s tests).

## Notes
- Adds unit tests for the bootstrap readiness helper (incl. the timeout diagnostic).
- No production behavior change beyond the bootstrap readiness gate (which is strictly more correct — exec'ing a not-ready container can only fail).

### DIFF
--- a/crate/operator/bootstrap.py
+++ b/crate/operator/bootstrap.py
@@ -204,16 +204,8 @@ async def _bootstrap_user_via_sidecar(
             raise _temporary_error()
 
 
-# How long a single bootstrap attempt waits for the crate container to become
-# ready before deferring back to kopf's handler retry. Polls frequently so the
-# exec fires the moment the container is up, rather than bouncing off a
-# not-ready container with 400/500s and waiting a full retry cycle.
-#
-# Generous on purpose: a crate container's readiness covers disk attach (slow on
-# azure-disk), init containers (incl. a JMX-jar download), JVM boot and the
-# readiness probe -- routinely a few minutes under load, so a tight budget would
-# just churn kopf retries. This is only a per-attempt cap; kopf still retries
-# beyond it, bounded by the overall BOOTSTRAP_TIMEOUT.
+# Per-attempt cap. readiness can take minutes under load (disk attach, init
+# containers, JVM boot), and kopf retries beyond it up to BOOTSTRAP_TIMEOUT.
 _READINESS_POLL_INTERVAL = 2
 _READINESS_MAX_WAIT = 300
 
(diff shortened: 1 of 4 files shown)


crate/operator/bootstrap.py
```diff
@@ -266,12 +395,24 @@ async def pod_exec(cmd):
                 tty=False,
             )
 
+    # Don't exec into a container that isn't ready yet: that yields a
+    # 400/500 'Invalid response status' and a wasted retry cycle. Wait for the
+    # crate container to be ready first; if it isn't within this attempt's
+    # budget, defer to kopf's handler retry.
```
tomach: The docstring of `def _wait_for_crate_container_ready` basically already explains the 400/500 problem, so this comment is pure restatement, can we drop it?

crate/operator/bootstrap.py
```diff
@@ -203,6 +204,134 @@ async def _bootstrap_user_via_sidecar(
...
+# How long a single bootstrap attempt waits for the crate container to become
+# ready before deferring back to kopf's handler retry. Polls frequently so the
+# exec fires the moment the container is up, rather than bouncing off a
+# not-ready container with 400/500s and waiting a full retry cycle.
+#
+# Generous on purpose: a crate container's readiness covers disk attach (slow on
+# azure-disk), init containers (incl. a JMX-jar download), JVM boot and the
+# readiness probe -- routinely a few minutes under load, so a tight budget would
+# just churn kopf retries. This is only a per-attempt cap; kopf still retries
+# beyond it, bounded by the overall BOOTSTRAP_TIMEOUT.
```
tomach:
```suggestion
# Per-attempt cap. readiness can take minutes under load (disk attach, init
# containers, JVM boot), and kopf retries beyond it up to BOOTSTRAP_TIMEOUT.
```

tomach (approved):
💯 looks great, thank you! but the inline commenting is a bit heavy sometimes, so the code becomes hard to read. for the future, can we pls:
- keep the explanatory "why" in the function docstrings, and
- trim the inline comments to 1–2 lines, only where the code can't speak for itself?

PR #845, opened 2026-06-16
First-class dedicated master node support

WalBeh:
ref:  https://github.com/crate/cloud/issues/2744

Makes dedicated master nodes (`spec.nodes.master`) behave consistently across all
cluster operations, with the invariant that master nodes are available before data
nodes (cluster formation/discovery depends on it).

## Changes
- **Node-group enumerator** (`iter_node_groups`) — single source of truth for
  iterating master + data groups, masters first.
- **Suspend / resume** — masters start before data on resume and stop after data on
  suspend; a standalone `master.replicas -> 0` while data is up is rejected.
- **Change-compute** — applied per node group (incl. master), with master resources
  read independently; `-Cprocessors` updated on the master StatefulSet.
- **Expand-volume** — master disk expansion handled alongside data groups.
- **Health ping** — robust node lookup (no longer assumes a `hot` data group).
- **Webhooks** — scale reports effective master replicas (`0` while suspended),
  change-compute includes master compute, expand-storage reports per-group disk size
  (all additive / backward-compatible).

## Tests
- Unit tests across the above (passing).
- e2e suite (`tests/test_master_nodes.py`, minikube-only, skipped without
  `--kube-config`): cluster creation with dedicated masters validated end-to-end on
  minikube; the full suspend/resume/compute/expand suite is intended to run on a
  cluster with more headroom (CI / dev).

## Follow-ups (out of scope here)
- cloud-api companion PR to consume the new webhook fields and define master-node
  products.
- Adding masters to an existing master-less cluster is **not** supported (requires a
  full restart).

PR #848, opened 2026-06-23
Stabilise e2e exposure tests: widen waits + timeout diagnostics

WalBeh:
Generic e2e/test-stability fixes extracted from the dedicated-master-nodes branch (#845) so master gets them without waiting for that feature. None of these touch master-node logic — they only change `tests/test_exposure.py` and the shared `tests/utils.py` helper. Same spirit as the already-merged #847.

## 1. Stabilise `test_change_exposure_*` / `test_create_traefik_*`
The exposure tests waited only `DEFAULT_TIMEOUT` (60s) for `cluster_create`. During bootstrap the operator execs into the pod over a websocket that can transiently return 400 until the container is ready; the operator retries, but 60s wasn't enough to absorb the retries, so `cluster_create` occasionally timed out (flaky, unrelated to master nodes). Wait `DEFAULT_TIMEOUT * 5`, matching `start_cluster`'s healthy-wait.

## 2. Widen cluster-create/bootstrap waits to 600s under parallel CI
A nightly run showed `test_change_exposure_loadbalancer_to_traefik` time out at 300s while the pod's diagnostic state was fully healthy. The cause isn't a stuck pod or retry cadence — total cluster bring-up (image pulls, JVM start, node contention) on a busy 3-node cluster running clusters in parallel simply exceeded the 300s budget. Bump the *initial* cluster-create/bootstrap waits (`start_cluster`'s bring-up waits and the inline `cluster_create` waits in `test_exposure`) from `*5` to `*10`. This only raises the ceiling — the happy path returns as soon as the handler finishes — and stays well under the operator's 1800s `BOOTSTRAP_TIMEOUT`. Post-create operation waits (`cluster_update`) are left at `*5`.

## 3. Report pod states when `wait_for_kopf_handler` times out
The wait already polls the full timeout to let the cluster come up; when it still times out (most often a stuck bootstrap), raise with each pod's container state instead of a bare "did not finish within Ns". Surfaces the actionable reason — `ImagePullBackOff`, `CrashLoopBackOff`, `ErrImagePull`, `CreateContainerError` — and falls back to pod conditions when no container exists yet (the `Pending` case), so a CI failure explains itself.

## Notes
- Test-only changes; no production behaviour change.
- GitHub Actions skips k8s tests, so these run on the nightly Jenkins integration suite.

LLM-assisted.

### DIFF
--- a/tests/test_exposure.py
+++ b/tests/test_exposure.py
@@ -162,10 +162,11 @@ async def test_create_traefik_with_cidrs(
         name,
         namespace.metadata.name,
         f"{KOPF_STATE_STORE_PREFIX}/cluster_create",
-        # Bootstrap execs into the pod over a websocket that can transiently
-        # 400 until the container is ready; the operator retries, so allow
-        # enough time to absorb those retries (matches start_cluster's *5).
-        timeout=DEFAULT_TIMEOUT * 5,
+        # Provisioning and bootstrapping a cluster on a busy parallel CI
+        # cluster can take several minutes (image pulls, JVM start, node
+        # contention), so give the wait generous headroom -- the operator
+        # already retries the bootstrap exec every few seconds in testing.
+        timeout=DEFAULT_TIMEOUT * 10,
     )
 
     # Service should be ClusterIP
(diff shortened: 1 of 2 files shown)


tomach (approved):
👍

PR #849, opened 2026-06-26
Add separate spec.grandCentral.exposure switch to decouple GC routing from cluster exposure

tomach:
## Summary of changes
Adds an optional, independent switch `spec.grandCentral.exposure` (`nginx` | `traefik`) so grand-central routing can be migrated to Traefik while the CrateDB service stays on a LoadBalancer.

- Unset -> falls back to `spec.cluster.exposure`, so existing clusters behave identically
- Set -> fully decouples grand-central from the CrateDB service, e.g. `cluster.exposure: loadbalancer` + `grandCentral.exposure: traefik`.
- Considered on create, suspend/resume, exposure changes, and allowed-CIDR updates.

 New CRD field is optional with no `default`, so existing clusters are untouched and the update handler does not fire on operator deploy.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/2744

### DIFF
--- a/crate/operator/handlers/handle_update_allowed_cidrs.py
+++ b/crate/operator/handlers/handle_update_allowed_cidrs.py
@@ -28,6 +28,7 @@ from kubernetes_asyncio.client import CoreV1Api, NetworkingV1Api
 from crate.operator.constants import GRAND_CENTRAL_RESOURCE_PREFIX
 from crate.operator.exposure import update_traefik_ip_restriction
 from crate.operator.grand_central import (
+    grand_central_uses_traefik,
     read_grand_central_httproute,
     read_grand_central_ingress,
     update_grand_central_ip_allowlist,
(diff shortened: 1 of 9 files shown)


plaharanne (approved):
looks good to me

crate/operator/handlers/handle_update_allowed_cidrs.py
```diff
@@ -92,6 +93,9 @@ async def update_service_allowed_cidrs(
         if exposure == "traefik":
             await update_traefik_ip_restriction(namespace, name, new_cidrs, logger)
 
+        # Grand-central IP allowlist follows its own exposure, which
+        # may differ from the CrateDB service exposure.
+        if grand_central_uses_traefik(cratedb["spec"]):
```
juanpardo, line 98: I think you may be missing the opposite case: when the cluster uses traefik but grand-central is on ingress

tomach: no, it's covered - the else belongs to `grand_central_uses_traefik(...)`, not the crate exposure check above (that one only patches the crate-service middleware and has no else). So `cluster.exposure: traefik` + `grandCentral.exposure: nginx` resolves to False -> else -> patches the GC nginx Ingress whitelist. I added some clarifying comments in https://github.com/crate/crate-operator/pull/849/changes/ede44d32dd1eac217139a7a086cae3ef63b70420

juanpardo (approved):

PR #864, opened 2026-07-20
Make legacy resume test assert data-node readiness, not LB health

tomach:
## Summary of changes
The end-to-end health check reaches CrateDB through the Azure load balancer, which is deleted on suspend and re-provisioned on resume, its new IP is often not routable yet, making the check flaky. Assert the data StatefulSet scales back up instead.

## Checklist

- [ ] Link to issue this PR refers to:


plaharanne (approved):
👍

PR #868, opened 2026-07-23
StackIT support

goat-ssh:
## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/3025
- [x] Relevant changes are reflected in `CHANGES.rst`
- [x] Added or changed code is covered by tests
- [x] Documentation has been updated if necessary
- [x] Changed code does not contain any breaking changes (or this is a major version change)

PR #869, opened 2026-07-28
Add STACKIT (SKE) support

WalBeh:
## Summary of changes
Adds STACKIT Kubernetes Engine (SKE) as a `CLOUD_PROVIDER` value. SKE hands pods carrier-grade NAT addresses out of `100.64.0.0/10`, which are not site-local, so CrateDB's own default of `network.host: _site_` finds no candidate address and the node refuses to start - bootstrapping on SKE fails outright:

```
No up-and-running site-local (private) addresses found, got [name:lo (lo), name:eth0 (eth0)]
```

Bind on `0.0.0.0` inside the pod's network namespace and publish the pod IP from the downward API instead. Also sets `node.attr.zone` from the OpenStack metadata service and spreads pods across availability zones, as on the other providers. One commit per provider branch.

Every branch is gated on `CLOUD_PROVIDER == stackit`, and `POD_IP` is appended last in `get_statefulset_crate_env()`, so rendered pod specs on aws/azure/gcp/openshift are unchanged - a StatefulSet spec change would roll-restart every managed cluster. Access is not widened either: the `crate` trust rule stays pinned to `_local_`. Both are covered by tests.

`TestStackitZoneAttribute` swaps `curl` for `cat` and runs the generated zone snippet through `sh`, because a silently empty extraction is the realistic failure mode and a string assertion would not catch it. Confirmed on an SKE dev cluster that `169.254.169.254` is reachable from inside a pod.

Zone names are `eu01-1`, `eu01-2`, ... not the `eu-central-1a` style of the existing AWS example. They are typed by hand into `cluster.routing.allocation.awareness.force.zone.values`, and a mismatch silently stops CrateDB forcing shard copies apart, so they are documented.

Not included: the `lb.stackit.cloud/*` service annotations (crate/cloud#2926 flags the idle-timeout behaviour as unverified, so the values would be guesses), and restore from STACKIT object storage (scheduled backups already work through `spec.backups.aws.endpointUrl`, but `AwsBackupRepositoryData` has no endpoint field). No e2e run against SKE yet. The PDB naming bug found while planning this is crate/cloud#3037, independent of this PR.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/3036

### DIFF
--- a/crate/operator/create.py
+++ b/crate/operator/create.py
@@ -57,6 +57,7 @@ from kubernetes_asyncio.client import (
     V1NodeSelector,
     V1NodeSelectorRequirement,
     V1NodeSelectorTerm,
+    V1ObjectFieldSelector,
     V1ObjectMeta,
     V1OwnerReference,
     V1PersistentVolumeClaim,
@@ -578,6 +579,16 @@ def get_statefulset_crate_command(
             }
         )
 
+    # STACKIT (SKE) assigns pods carrier-grade NAT addresses (100.64.0.0/10). Those
+    # are not site-local, so CrateDB's default ``_site_`` host resolution finds no
+    # candidate address and the node refuses to start (crate/cloud#2926). Bind on all
+    # interfaces of the pod's own network namespace and publish the pod IP instead.
+    # This does not widen access: the ``crate`` trust rule stays pinned to ``_local_``
+    # and every other connection keeps going through password or JWT auth.
+    if config.CLOUD_PROVIDER == CloudProvider.STACKIT:
+        settings["-Cnetwork.host"] = "0.0.0.0"
+        settings["-Cnetwork.publish_host"] = "$(POD_IP)"
+
     # Availability zone retrieval at pod launch time
     if config.CLOUD_PROVIDER == CloudProvider.AWS:
         aws_cmd = (
@@ -677,6 +688,21 @@ def get_statefulset_crate_env(
                     ),
                 ),
             ]
+        )
+
+    # Referenced as ``$(POD_IP)`` by ``-Cnetwork.publish_host`` on STACKIT. Appended
+    # only there, and last, so the pod spec of every other provider stays byte for
+    # byte what it was - an extra env var would restart all existing clusters.
+    if config.CLOUD_PROVIDER == CloudProvider.STACKIT:
+        crate_env.append(
+            V1EnvVar(
+                name="POD_IP",
+                value_from=V1EnvVarSource(
+                    field_ref=V1ObjectFieldSelector(
+                        api_version="v1", field_path="status.podIP"
+                    )
+                ),
+            )
         )
 
     return crate_env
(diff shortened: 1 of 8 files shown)


crate/operator/create.py
```diff
@@ -597,6 +609,16 @@ def get_statefulset_crate_command(
         settings["-Cnode.attr.zone"] = (
             f"$(curl -s '{url}' -H 'Metadata-Flavor: Google' | awk -F'/' '{{print $NF}}')"  # noqa
         )
+    elif config.CLOUD_PROVIDER == CloudProvider.STACKIT:
+        url = "http://169.254.169.254/openstack/latest/meta_data.json"
```
goat-ssh: Maybe cleaner with https://github.com/crate/crate-operator/pull/868/changes#diff-8dae7e9e202e83076ae11a5fde28970f59a3f42b8138a1fb22ea4b62d57d942eR602-R604

WalBeh: Good call — that would drop the whole `tr | grep | cut` pipeline, and it reuses the same path the AWS branch already hits, so it'd read more like the rest of the file.

The one thing I can't check myself: does `/latest/meta-data/placement/availability-zone` actually answer on SKE? All I have is the `openstack/latest/meta_data.json` output from a dev cluster, which is why I parsed that. OpenStack does implement the EC2-compat API, so it probably works — but I'd rather not swap a verified path for an unverified one here. If it 404s, `curl -s` hands CrateDB the error body as the zone, every node gets the same bogus value, and forced zone allocation silently stops separating shard copies. That's the failure mode the executed-pipeline test in `tests/test_stackit.py` is there to catch.

Did you run #868 against an actual SKE cluster? If so — or if anyone can paste a `curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone` from a pod — I'll switch it over and delete the pipeline and its test happily.

(Unrelated, since your linked hunk covers it too: I went with `$(POD_IP)` rather than `${POD_IP}` so the kubelet does the substitution, instead of relying on the entrypoint's shell.)

goat-ssh: yes, i can confirm the `eu01-1 eu01-2 eu01-3` from the live cluster.

WalBeh:
```
curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone
eu01-3
```

WalBeh: Confirmed on a dev cluster — the endpoint returns a bare `eu01-3`, no newline, nothing to parse. Switched to it in dad4ac9, so the `tr | grep | cut` pipeline and the test that executed it are both gone. Net -43 lines. Thanks for the pointer.

One wrinkle worth knowing if you touch these tests: the negative test can no longer assert "this provider does not mention the metadata URL", because AWS reads the *same* path behind its IMDSv2 token. It now compares the whole `-Cnode.attr.zone=$(curl -s '...')` snippet instead.

plaharanne (approved):
👍

goat-ssh (approved):

PR #870, opened 2026-07-28
Give every node spec its own PodDisruptionBudget

WalBeh:
## Summary of changes
`create_statefulset()` runs once per node spec - once for `spec.nodes.master` and once per entry in `spec.nodes.data` - but named the `PodDisruptionBudget` `crate-<name>` every time, while only the selector varied by `node-name`. Since the create passes `continue_on_conflict=True`, every call after the first got a 409 that `call_kubeapi()` turns into a warning and swallows.

So a cluster ended up with one budget covering only the node spec registered first:

| cluster | result |
| --- | --- |
| one data group, no masters | one budget, correct - why this went unnoticed |
| dedicated master nodes | budget covers the **masters**, data nodes have none |
| several data groups | only the first group is covered |

The dedicated-master case is the bad one: it protects the nodes that matter least for shard availability and leaves the data nodes free to be drained without limit.

Fix is to name the budget after the `StatefulSet` it guards (`crate-master-<name>`, `crate-data-hot-<name>`, ...), so the names no longer collide. The selector already carried `node-name`, and the existing comment states the intent was one budget per StatefulSet, so this restores the intended behaviour rather than redefining it. Extracted into `get_pod_disruption_budget()` to make it unit-testable.

Verified the tests are a real regression guard: reverting just the name expression back to `crate-<name>` fails 4 of the 5 new tests.

### Existing clusters - no backfill needed

Budgets are only created at cluster creation and nothing reconciles them afterwards, so clusters created before this change keep whatever they got. That turns out not to matter:

- A cluster with a single data group and no dedicated masters gets exactly one budget, correctly scoped. That is the shape of every cluster in use.
- The two broken shapes are not in use. Dedicated master nodes only shipped in 2.62.0 and are not selectable from the console yet. Multiple data groups are structurally allowed by the CRD but not actually supported - six code paths read `nodes["data"][0]` directly and no test exercises a second group.

Worth a quick check that no master-enabled cluster was created by hand since 2.62.0, but otherwise there is nothing to migrate.

Two things I deliberately did not do, both worth a second opinion:

- **No cluster-wide budget.** A single budget selecting all pods of a cluster would need no rename and no migration, but it serialises disruption across masters and every data group, making node pool upgrades much slower on multi-group clusters. It is also a semantic change rather than a bug fix.
- **Per-group budgets do allow two data pods to be evicted at once** on a multi-group cluster (one per group). If a table's primary and replica happen to sit on nodes in different groups, that is a shard availability risk. This PR does not change that behaviour - it is how the selector was always meant to work - but if we would rather have a cluster-wide limit, say so and I will switch it.

No e2e coverage: the k8s suite would need a dedicated-master cluster to assert budget coverage per group.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/3037

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,13 @@ Changelog
 Unreleased
 ----------
 
+* Fixed the ``PodDisruptionBudget`` being created once per node spec but always
+  named after the cluster. Only the first node spec ended up with a budget, the
+  rest were dropped as name conflicts - on clusters with dedicated master nodes
+  that left the data nodes unprotected during node drains. Budgets are now named
+  after the ``StatefulSet`` they guard. Clusters created before this release keep
+  their single budget until they are recreated.
+
 2.62.2 (2026-07-22)
 -------------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,12 +5,7 @@ Changelog
 Unreleased
 ----------
 
-* Fixed the ``PodDisruptionBudget`` being created once per node spec but always
-  named after the cluster. Only the first node spec ended up with a budget, the
-  rest were dropped as name conflicts - on clusters with dedicated master nodes
-  that left the data nodes unprotected during node drains. Budgets are now named
-  after the ``StatefulSet`` they guard. Clusters created before this release keep
-  their single budget until they are recreated.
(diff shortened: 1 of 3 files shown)


plaharanne (approved):
Looks good to me

PR #871, opened 2026-07-30
Support clusters with more than one data node group

WalBeh:
## Summary of changes

Four bugs made a cluster with two or more data node groups unusable — three from #3038 plus one this branch initially reintroduced (see the second commit). They go together: the first is what currently stops the others from being reached, so fixing it alone would trade a loud retry loop for silent failures.

**Every update raised `KeyError`.** `update_cratedb()` rebound the loop variable it was iterating, so the second group's lookup indexed a dict with an integer. It threw before any sub-handler was registered, so scale, suspend, resume, disk expansion and compute change were all dead. Now iterates `(old, new)` pairs under their own names.

**Suspend deadlocked.** `check_cluster_healthy()` and `check_all_data_nodes_gone()` were called inside the per-group loop, and both deadlock there: a half-suspended cluster is never GREEN, and waiting for *every* data pod to be gone cannot succeed while the groups behind the current one are still running. The StatefulSets are now collected first so both run once around the whole set, with the health gate skipped once anything is already down — which is also what keeps the retry after the wait from deadlocking in the gate instead. (That signal cannot distinguish a retry from a StatefulSet scaled down out of band; noted in the code.)

**Scale-down leaked allocation exclusions.** `excess_nodes` was overwritten per group, so `reset_allocation()` cleared only the last one. The rest stayed excluded in the persistent cluster state, and a later scale-up returned those nodes to the cluster holding no shards. Now accumulated across groups, while `deallocate_nodes()` still gets only the group it is removing.

**A master scale to zero could be waved through — found in review of the first commit, and caused by it.** Suspend/resume was being decided in two places with different rules: the data dispatch asked whether *every* group crosses zero (matching `scale.classify_data_scaling`), while `_data_nodes_cross_zero` asked whether *any* group does — and that one gates `master.replicas`. On a two-group cluster, `master.replicas: 3 → 0` together with `cold: 2 → 0` passed the guard, then `classify_data_scaling` routed it to `scale_cluster` rather than `suspend_or_start_cluster`, which honoured the master change literally and took **every master down while `hot` was still serving**. Before this branch the `KeyError` aborted the update first, so the first commit is what made it reachable. Replaced with a single `_classify_data_scaling()` returning `(resuming, suspending)`, used by both sites so they cannot drift again.

**Adding or removing a data group is now rejected** instead of silently ignored. `zip` truncates, so adding a group as the only change registered no sub-handler at all and the handler returned early: no error, no resources, and the CRD and the cluster quietly diverged. `ScaleSubHandler` already rejects this but never gets registered from here. Not a regression — it behaved the same before this branch — but it is a silent failure in code this branch is already fixing, and it answers one of the open questions on #3038.

Suspend and resume are now decided from every group rather than per group throughout. Without that, one group of several going to zero counted as a suspend in the dispatch while the scale handler performed an ordinary scale, and the mismatch skipped `after_cluster_update`, leaving `cluster.routing.allocation.enable` pinned to `new_primaries`.

Tests sit next to each module's existing ones — `tests/test_handle_update_cratedb.py` (new), `TestSuspendWithSeveralDataGroups` in `tests/test_operations.py`, `TestScaleDownAllocationReset` in `tests/test_scale.py`. There was no two-data-group fixture anywhere before this. Each is a real regression guard, confirmed by restoring the old behaviour and watching them fail.

No backfill needed — there is no cluster with more than one data node group today, so nothing is carrying leaked exclusions.

### Follow-ups this surfaced, deliberately not in scope

All three need a multi-group cluster to manifest, so none can bite today. The first two change what the webhook consumer receives, so they want agreement from that side rather than a unilateral edit here:

- **`change_compute.py:148`** builds the `COMPUTE_CHANGED` payload from `data[0]`. A compute change on a later group now dispatches correctly (it used to `KeyError`), but the payload carries no delta, so billing sees a compute change that changed nothing. `iter_changed_compute_groups` is the right source; the payload shape needs a group.
- **`webhooks.py` `WebhookAction.for_diff`** still labels a single group crossing zero as `suspend`, so a failed partial scale-down sends a failure notification labelled `suspend`. Same `any`/`all` mismatch, cosmetic.
- **`operation` is overwritten by later groups**, so a diff that scales one group and changes another's compute ends up labelled by whichever comes last in the spec. Picking a precedence is a design call, and it sits with the `if/elif`-chain limitation already listed on #3038.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/3038
- [ ] Documentation has been updated if necessary (no doc describes this behaviour)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,12 @@ Unreleased
 
 * Fixed rolling restart progress restarting from 1 for every node group.
 
+* Fixed every update failing on clusters with more than one data node group.
+
+* Fixed suspending a cluster with more than one data node group getting stuck.
+
+* Fixed stale shard allocation exclusions after scaling down several data node groups.
+
 2.62.2 (2026-07-22)
 -------------------
 
--- a/crate/operator/handlers/handle_update_cratedb.py
+++ b/crate/operator/handlers/handle_update_cratedb.py
@@ -22,7 +22,7 @@
 import datetime
 import hashlib
 import logging
-from typing import List
+from typing import List, Tuple
 
 import kopf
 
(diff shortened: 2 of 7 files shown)


tomach (approved):
👍

PR #872, opened 2026-07-30
Validate data node group names, and count restart progress properly

WalBeh:
## Summary of changes

Two unrelated node-group bugs, both cheap, from the same review as #3037.

**Data node group names are now validated.** `validate_node_spec()` rejects names Kubernetes will not accept, the reserved `master` name, and names that repeat within a cluster. The CRD schema is looser than Kubernetes on all of these, and none of the shapes it lets through can actually start:

- an underscore, a capital, a trailing newline, or more than 63 characters produces a StatefulSet the API server rejects with an opaque 422;
- two groups sharing a name produce the **same** StatefulSet name, where the second create is a 409 that `call_kubeapi` swallows — so one group silently gets no StatefulSet while its replicas still count towards `gateway.expected_data_nodes`, and the cluster never forms;
- a group named `master` gets the same `node-name` label as the dedicated master group, so the two StatefulSets end up with **identical selectors** and fight over each other's pods — and it collapses the `{name: group}` dict in `iter_changed_compute_groups`, so a master-only compute change is diffed against the wrong spec.

The 63-character bound covers the `node-name` label value, which is the name verbatim. It does not bound the composite StatefulSet name, which also carries the cluster name — `validate_node_spec()` does not receive that, so it stays unbounded as before.

**Rolling restart progress counts properly.** It reported the pod's ordinal as a cluster-wide index, but ordinals restart at 0 in every StatefulSet, so restarting a cluster with dedicated masters sent `1/6, 2/6, 3/6, 1/6, 2/6, 3/6` to the webhook consumer. It now counts off the pods still to do, against what *this* run set out to restart — which for a compute change is only the changed groups, not every pod in the cluster.

That needed the run's initial pod count in a new `status.pendingPodsTotal`, alongside the existing `pendingPods`. A restart already in flight when this ships has no total recorded, so it is seeded from the remaining queue and **persisted** — re-deriving it each pass would track the shrinking queue and report `1/5, 1/4, 1/3`, with the numerator stuck and the denominator going backwards. It is cleared with `pendingPods` on both completion paths and in `rollback.py`, so the two never outlive each other.

### The CRD schema is deliberately left alone

The issue proposed tightening the `name` pattern and adding `uniqueItems`. I did neither:

- **Tightening the pattern** would also validate every future write to CrateDB resources that *already exist*, so a cluster carrying a name the new pattern rejects could no longer be updated at all. Whether such a cluster exists is still the open question on #3039, and the blast radius is much larger than the check added here.
- **`uniqueItems` would not help.** It compares whole items, so two groups named the same but differing in `replicas` are distinct values and stay allowed.

`validate_node_spec()` runs on create only (`handle_create_cratedb.py`), so this closes the hole for new clusters and cannot affect a running one. For the same reason it is not called from the update handler: that would block every subsequent update to a cluster whose name is already bad. The CRD question stays open on the issue.

Not covered by a test: the one-line `pendingPodsTotal` clear in `rollback.py`. No existing test exercises that path and standing up the rollback machinery for a status-hygiene line looked disproportionate — say the word and I will add it.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/3039
- [ ] Documentation has been updated if necessary (no doc describes CRD field constraints)

### DIFF
--- a/crate/operator/operations.py
+++ b/crate/operator/operations.py
@@ -156,8 +156,22 @@ def get_master_nodes_names(nodes: Dict[str, Any]) -> List[str]:
 #: A data node group's name ends up in the StatefulSet name
 #: (``crate-data-<group>-<cluster>``) and in the CrateDB node names
 #: (``data-<group>-<i>``), so it has to be a valid Kubernetes name fragment:
-#: lowercase alphanumeric and dashes, starting and ending alphanumeric.
-DATA_NODE_GROUP_NAME = re.compile(r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?$")
+#: lowercase alphanumeric and dashes, starting and ending alphanumeric. Matched
+#: with ``fullmatch`` -- with ``match``, ``$`` would also accept a trailing
+#: newline, which is not a valid Kubernetes name.
+DATA_NODE_GROUP_NAME = re.compile(r"[a-z0-9]([-a-z0-9]*[a-z0-9])?")
+
+#: The name also becomes the ``node-name`` label value verbatim, and Kubernetes
+#: caps label values at 63 characters. This does not bound the *composite*
+#: StatefulSet name, which also carries the cluster name -- that needs the
+#: cluster name, which this validation does not get (crate/cloud#3039).
+DATA_NODE_GROUP_NAME_MAX_LENGTH = 63
+
+#: ``master`` is reserved for the dedicated master group, which is created with
+#: ``node_name="master"``. A data group of that name would get the same
+#: ``node-name`` label, so its StatefulSet selector would be identical to the
+#: master StatefulSet's and the two would fight over each other's pods.
+RESERVED_NODE_GROUP_NAMES = frozenset({"master"})
 
 
 def validate_node_spec(nodes: Dict[str, Any], logger: logging.Logger) -> None:
@@ -172,14 +186,20 @@ def validate_node_spec(nodes: Dict[str, Any], logger: logging.Logger) -> None:
     are configured their replica count must be odd and at least three, so the
     masters can always form a quorum.
 
-    Data node group names must be valid Kubernetes name fragments and unique
-    within the cluster. The CRD schema is looser than Kubernetes itself on both
-    counts, and neither shape can start (crate/cloud#3039): a name with an
-    underscore or a capital produces a StatefulSet the API server rejects with an
-    opaque 422, and two groups sharing a name produce the *same* StatefulSet
-    name, where the second create is a 409 that ``call_kubeapi`` swallows -- so
-    one group silently gets no StatefulSet while its replicas still count towards
-    ``gateway.expected_data_nodes``, and the cluster never forms.
+    Data node group names must be valid Kubernetes name fragments, not clash with
+    the reserved ``master`` group, and be unique within the cluster. The CRD
+    schema is looser than Kubernetes itself, and none of the shapes it lets
+    through can start (crate/cloud#3039):
+
+    - a name with an underscore or a capital, or one over 63 characters,
+      produces a StatefulSet the API server rejects with an opaque 422;
+    - two groups sharing a name produce the *same* StatefulSet name, where the
+      second create is a 409 that ``call_kubeapi`` swallows -- so one group
+      silently gets no StatefulSet while its replicas still count towards
+      ``gateway.expected_data_nodes``, and the cluster never forms;
+    - a group named ``master`` gets the same ``node-name`` label as the
+      dedicated master group, so the two StatefulSets end up with identical
+      selectors and fight over each other's pods.
 
     :param nodes: The ``spec.nodes`` from a CrateDB custom resource.
     :param logger: Logger used to record why the spec was rejected.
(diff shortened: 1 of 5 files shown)


crate/operator/operations.py
```diff
@@ -152,6 +153,27 @@ def get_master_nodes_names(nodes: Dict[str, Any]) -> List[str]:
         return [f"data-{node_name}-{i}" for i in range(node["replicas"])]
 
 
+#: A data node group's name ends up in the StatefulSet name
+#: (``crate-data-<group>-<cluster>``) and in the CrateDB node names
+#: (``data-<group>-<i>``), so it has to be a valid Kubernetes name fragment:
+#: lowercase alphanumeric and dashes, starting and ending alphanumeric. Matched
+#: with ``fullmatch`` -- with ``match``, ``$`` would also accept a trailing
+#: newline, which is not a valid Kubernetes name.
```
tomach:
```suggestion
#: Valid Kubernetes name fragment; ``fullmatch`` so a trailing newline is
#: rejected. The name ends up in the StatefulSet name and CrateDB node names.
```

crate/operator/operations.py
```diff
@@ -152,6 +153,27 @@ def get_master_nodes_names(nodes: Dict[str, Any]) -> List[str]:
...
+DATA_NODE_GROUP_NAME = re.compile(r"[a-z0-9]([-a-z0-9]*[a-z0-9])?")
+
+#: The name also becomes the ``node-name`` label value verbatim, and Kubernetes
+#: caps label values at 63 characters. This does not bound the *composite*
+#: StatefulSet name, which also carries the cluster name -- that needs the
+#: cluster name, which this validation does not get (crate/cloud#3039).
```
tomach:
```suggestion
#: The name is the ``node-name`` label value verbatim; Kubernetes caps those
#: at 63. Does not bound the composite StatefulSet name (crate/cloud#3039).
```

crate/operator/operations.py
```diff
@@ -152,6 +153,27 @@ def get_master_nodes_names(nodes: Dict[str, Any]) -> List[str]:
...
+DATA_NODE_GROUP_NAME_MAX_LENGTH = 63
+
+#: ``master`` is reserved for the dedicated master group, which is created with
+#: ``node_name="master"``. A data group of that name would get the same
+#: ``node-name`` label, so its StatefulSet selector would be identical to the
+#: master StatefulSet's and the two would fight over each other's pods.
```
tomach:
```suggestion
#: ``master`` is the dedicated master group's node name; a data group of that
#: name would share its selector and the two would fight over each other's pods.
```

crate/operator/operations.py
```diff
@@ -164,6 +186,21 @@ def validate_node_spec(nodes: Dict[str, Any], logger: logging.Logger) -> None:
...
+    - a name with an underscore or a capital, or one over 63 characters,
+      produces a StatefulSet the API server rejects with an opaque 422;
+    - two groups sharing a name produce the *same* StatefulSet name, where the
+      second create is a 409 that ``call_kubeapi`` swallows -- so one group
+      silently gets no StatefulSet while its replicas still count towards
+      ``gateway.expected_data_nodes``, and the cluster never forms;
+    - a group named ``master`` gets the same ``node-name`` label as the
+      dedicated master group, so the two StatefulSets end up with identical
+      selectors and fight over each other's pods.
+
```
tomach:
```suggestion
    Data node group names must be valid Kubernetes name fragments, not clash
    with the reserved ``master`` group, and be unique within the cluster. The
    CRD schema is looser than Kubernetes on all three, and none of the shapes it
    lets through can start (crate/cloud#3039).
```

crate/operator/operations.py
```diff
@@ -741,27 +820,41 @@ async def restart_cluster(
...
     next_pod_uid = pending_pods[0]["uid"]
     next_pod_name = pending_pods[0]["name"]
 
+    # Progress is counted off the pods still to do, not the pod's ordinal: those
+    # restart at 0 in every StatefulSet, so a masters + data restart reported
+    # 1/6, 2/6, 3/6, 1/6, ... (crate/cloud#3039). The total is what *this* run set
+    # out to restart, which for a compute change is only the changed groups. A
+    # restart already in flight when this field was introduced has none recorded,
+    # so seed it from what is left and persist it - re-deriving it every pass
+    # would track the shrinking queue and report 1/5, 1/4, 1/3, ...
```
tomach:
```suggestion
    # Count off the pods still to do, not the pod's ordinal (which restarts at 0
    # per StatefulSet). Seed and persist a total for a restart already in flight
    # when this field shipped; re-deriving it would track the shrinking queue.
```

tomach (approved):
LGTM! 💯 just a few suggestions to make the code more readable, if you feel like it.

crate/operator/operations.py
WalBeh: Applied. Two small deviations worth mentioning:

- Kept the blank line before the `:param` field list — without it Sphinx stops treating the params as a field list.
- Moved the duplicate-name failure mode down to a comment on the duplicate check itself. The other two now have their own constant to live next to, but the swallowed 409 is the one that no constant documents, and it is the least obvious of the three.

crate/operator/operations.py
WalBeh: Applied, with the `crate/cloud#3039` reference kept on the end — the rest of the file cites issues from comments the same way (e.g. `_lb_annotations_to_remove`), and it is the trail back to why the field exists at all.

Issue #863, opened 2026-07-17
Align `cluster.routing.allocation.enable` with CrateDB recommendation to use `primaries`


hammerhead:
The operator currently sets `cluster.routing.allocation.enable` to `new_primaries` before terminating nodes and therefore also during rolling upgrades:

https://github.com/crate/crate-operator/blob/4a45724b33ab1d486c7ae99795c8014ef60df777/crate/operator/operations.py#L777-L783

We recently found that using `new_primaries` across a rolling upgrade can cause an existing primary shard to become unavailable. The CrateDB documentation has therefore been changed to recommend `primaries` instead (https://github.com/crate/crate/issues/19751).

The operator is not directly affected by the reproduced scenario. It sets/resets the allocation setting before/after each individual node restart rather than at the beginning/end of the complete upgrade workflow.

Nevertheless, I suggest changing the operator to use `primaries` as well. `primaries` provides the same relevant behavior for the operator's workflow, while eliminating two different variants of the rolling upgrade process.

Otherwise, there is a risk that `new_primaries` is perceived as a generally safe setting for rolling upgrades. If the same approach is later applied in another deployment scenario without the important condition that allocation is restored after every node restart, it can lead to primary shards becoming unavailable.

Standardising on `primaries` establishes a single canonical recommendation that works with both per-node allocation changes and allocation changes spanning the complete rolling upgrade. The additional restriction imposed by `new_primaries` does not appear necessary and could become problematic if the operator is ever changed to use a different restart approach.

## Proposed change

Replace: `value="new_primaries"`
With: `value="primaries"`

plaharanne:
@seut Could you prioritize it? It has been like this for years, we can't tell if this is a bug or a technical suggestion.

PR #880, opened 2026-08-07
Use primaries rather than new_primaries while nodes are down

WalBeh:
## Summary of changes

Switches `cluster.routing.allocation.enable` from `new_primaries` to `primaries` while nodes are
taken down, in both places that write it.

Both values suppress replica allocation, which is why the setting is held across a restart at all.
`new_primaries` additionally returns `NO` for `EXISTING_STORE` recovery, so a node that comes back
cannot recover its own primaries from local disk while the setting is in force. `primaries` permits
that, and is what the CrateDB rolling-upgrade docs now recommend (crate/crate#19751).

**The Go half is the one that lands.** dc_util writes the same setting from its preStop hook — later
than the operator, and as `TRANSIENT` rather than `PERSISTENT`. `Metadata` builds the effective
settings as persistent with transient on top, so dc_util's value applies to every restart outside
OpenShift. Changing `operations.py` alone, as the issue proposes, would have been inert. The Python
constant keeps the operator's two call sites in agreement with the Go default.

`new_primaries` stays a valid `dc-util-pre-stop-routing-allocation` label value — rejecting it would
silently flip clusters that set it explicitly over to the default at the same moment the default
changes.

No change to the fallback: `restart_cluster` still resets the setting when the cluster is unhealthy
after a pod returns, and `RESET GLOBAL` clears both levels. `primaries` only makes that path rarer,
since reaching GREEN no longer depends on dc_util's postStart reset landing.

`--min-availability PRIMARIES` in the preStop hook is `cluster.graceful_stop.min_availability`, a
different setting that shares the word. Untouched.

Rollout: the binary is fetched from the fileserver at hook runtime, so the dc_util change takes
effect on the next pod restart without an operator release.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/crate-operator/issues/863

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,8 @@ Unreleased
 
 * Fixed stale shard allocation exclusions after scaling down several data node groups.
 
+* Changed ``cluster.routing.allocation.enable`` during restarts to ``primaries``.
+
 2.62.2 (2026-07-22)
 -------------------
 
--- a/crate/operator/operations.py
+++ b/crate/operator/operations.py
@@ -165,6 +165,11 @@ DATA_NODE_GROUP_NAME_MAX_LENGTH = 63
 #: name would share its selector and the two would fight over each other's pods.
 RESERVED_NODE_GROUP_NAMES = frozenset({"master"})
 
+#: Held while nodes are down so replicas are not re-allocated; not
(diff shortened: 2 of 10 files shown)


hammerhead (approved):
Thank you, looks good as far as I can tell without being very familiar with the code

PR #884, opened 2026-08-11
Fix the GC exposure test that failed on a finished cluster_create handler

WalBeh:
## Summary of changes

`test_gc_change_exposure_loadbalancer_to_traefik` failed all three reruns in the last nightly integration run with `Handler 'operator.cloud.crate.io/cluster_create' did not finish within 300s` — against a healthy cluster. The operator log for the same run shows the opposite: every sub-handler (`sql_exporter_config`, `system_user`, `services`, `statefulset_data_hot`, `bootstrap`) succeeded on all three attempts, with no temporary failures, and `cluster_create` succeeded 24s after creation. The failure is in the wait helper, not the operator.

`wait_for_kopf_handler` polled the kopf progress annotation and treated the handler as done only once it had seen that annotation and then seen it disappear. kopf never writes a terminal `success` into it — it deletes every progress annotation of a handling cycle in a single patch:

```
Patching with: {'status': {'cluster_create': {}, ...,
  'cluster_create/bootstrap': {'success': True, ...}},
 'metadata': {'annotations': {
   'operator.cloud.crate.io/cluster_create': None,
   'operator.cloud.crate.io/cluster_create.bootstrap': None, ...}}}
```

So `_check_kopf_handler_status` could only ever return `False` (running) or `None` (absent); its `success` branch was unreachable. A wait that starts after the cycle ended sees only the absent annotation, never sets its `seen` flag, and times out on a handler that already succeeded.

This test hit it because it was the only one in the file leaving `wait_for_lb` at its default `True`, so `start_cluster` blocked on the load balancer address before the wait began. `cluster_create` finished and purged its annotations while the test was still waiting for that address. That is also why it is flaky rather than always red — it passes only when the load balancer wins the race against cluster bring-up.

Changes:

- `tests/utils.py`: an absent annotation is now resolved against the sub-handler records in the resource `status`, which kopf's purge leaves alone. This separates "already finished" from "not started yet", which the annotation alone cannot express. The wait also defaults to `CLUSTER_CREATE_TIMEOUT`, which #848 introduced for this class of flake but applied only to `tests/test_exposure.py`.
- `tests/test_create_grand_central.py`: pass `wait_for_lb=False`, as the four sibling tests already do. The test discards the return value and only inspects the Ingress and the HTTPRoute, so it never needs the address.
- `tests/test_kopf_handler_wait.py`: new unit tests, no cluster required.

With the status check disabled, `test_returns_when_handler_finished_before_first_poll` fails and the rest pass, so it pins this specific bug. Full non-k8s suite: 396 passed.

The k8s test itself has not been run against a cloud provider, so the end-to-end pass is unverified; the operator-side analysis comes from the nightly log.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/3056
- [ ] Relevant changes are reflected in `CHANGES.rst`: not applicable, tests only, as in #848

### DIFF
--- a/tests/test_create_grand_central.py
+++ b/tests/test_create_grand_central.py
@@ -499,6 +499,10 @@ async def test_gc_change_exposure_loadbalancer_to_traefik(
         coapi,
         1,
         wait_for_healthy=False,
+        # The test only looks at the Ingress and the HTTPRoute, so it does not
+        # need the load balancer address. Waiting for one costs minutes on a
+        # cloud provider and delays the handler wait below.
+        wait_for_lb=False,
         additional_cluster_spec={
             "externalDNS": f"{name}.example.com",
         },
@@ -508,7 +512,6 @@ async def test_gc_change_exposure_loadbalancer_to_traefik(
         name,
         namespace.metadata.name,
         f"{KOPF_STATE_STORE_PREFIX}/cluster_create",
-        timeout=DEFAULT_TIMEOUT * 5,
     )
     await _start_gc_on_cluster(coapi, name, namespace.metadata.name)
 
--- a/tests/test_kopf_handler_wait.py
+++ b/tests/test_kopf_handler_wait.py
@@ -1,0 +1,110 @@
+# CrateDB Kubernetes Operator
+#
+# Licensed to Crate.IO GmbH ("Crate") under one or more contributor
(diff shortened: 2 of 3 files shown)


tomach (approved):
👍

PR #887, opened 2026-08-17
Require stable cluster health before advancing a rolling restart

tomach:
## Summary of changes
The rolling-restart health gate (`is_cluster_healthy`) decided it was safe to kill the next node from a single snapshot of `sys.nodes` and `sys.health`. Node membership becomes visible before the master finishes recomputing cluster health, so there is a short window where the count is back and `sys.health` still reports a stale `GREEN`.

Now it requires the same healthy reading to hold across N consecutive checks before returning healthy. Configurable via `HEALTH_CHECK_STABILITY_CHECKS` (default 3) and `HEALTH_CHECK_STABILITY_DELAY` (default 2s)

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/support/issues/916

### DIFF
--- a/tests/test_cratedb.py
+++ b/tests/test_cratedb.py
@@ -293,8 +293,8 @@ async def test_is_cluster_healthy_requires_stable_readings():
     logger = mock.Mock()
     # First poll catches a stale GREEN, the second sees the cluster go YELLOW
     # as the master recomputes health. A single snapshot used to pass here and
-    # let the next node be terminated too early (crate/cloud#3062); requiring
-    # consecutive readings must now report the cluster as not healthy.
+    # let the next node be terminated too early; requiring consecutive readings
+    # must now report the cluster as not healthy.
     conn_factory = _conn_factory_for([(3, 1), (3, 2)])
     result = await is_cluster_healthy(
         conn_factory,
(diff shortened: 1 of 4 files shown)


tests/test_cratedb.py
```diff
@@ -197,6 +202,142 @@ async def test_get_healthiness(healthiness):
...
+        stability_delay=0,
+    )
+    assert result is expected
+
+
+async def test_is_cluster_healthy_requires_stable_readings():
+    logger = mock.Mock()
+    # First poll catches a stale GREEN, the second sees the cluster go YELLOW
+    # as the master recomputes health. A single snapshot used to pass here and
+    # let the next node be terminated too early (crate/cloud#3062); requiring
```
plaharanne: I would not mention the issue from cloud repo here

plaharanne (approved):
I've added one note about a comment but except that, it looks good to me.

tests/test_cratedb.py
tomach: Thanks! Fixed in https://github.com/crate/crate-operator/pull/887/changes/ab0e72b96f970a2f40095d19985d455c2d7093e8

PR #888, opened 2026-08-18
Fix grand-central hostname corruption when cluster name recurs in external DNS

tomach:
## Summary of changes
The grand-central hostname derivation used an unbounded `str.replace()` to insert the `.gc` label, so a cluster name that recurs later in its external DNS (e.g. `us-west` in region `us-west-2`) got rewritten everywhere, producing a broken hostname. It now replaces only the first (leading-label) occurrence, which is safe because `external_dns` is always `<cluster_name>.<region>.<domain>.`

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/3065

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changelog
 Unreleased
 ----------
 
+* Fixed the grand-central hostname being corrupted when the cluster name recurs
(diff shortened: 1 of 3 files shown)


plaharanne (approved):
👍

PR #889, opened 2026-08-23
Support restoring a snapshot from an AWS S3 bucket with an endpoint URL

plaharanne:
## Summary of changes

Support restoring a snapshot from an S3-compatible bucket with a custom endpoint URL.
Previously the endpoint_url was ignored.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/3049

### DIFF
--- a/crate/operator/restore_backup.py
+++ b/crate/operator/restore_backup.py
@@ -246,7 +246,10 @@ async def get_source_backup_repository_data(
                 f'Secret {secret_key_ref["name"]} could not be found.'
             )
         except KeyError:
-            raise kopf.PermanentError(f"Key {key} not found in secret.")
+            if BackupRepositoryData.is_optional(backup_provider, key):
+                data_dict[key] = ""
+            else:
+                raise kopf.PermanentError(f"Key {key} not found in secret.")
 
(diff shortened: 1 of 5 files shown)


crate/operator/restore_backup.py
```diff
@@ -246,7 +246,10 @@ async def get_source_backup_repository_data(
                 f'Secret {secret_key_ref["name"]} could not be found.'
             )
         except KeyError:
-            raise kopf.PermanentError(f"Key {key} not found in secret.")
+            if BackupRepositoryData.is_optional(backup_provider, key):
+                data_dict[key] = ""
```
juanpardo, line 250: Is the entry really needed at all?
If so, perhaps a `None` would make more sense than an empty string, if it's allowed.
Otherwise, I would not add the key.

juanpardo (approved):
Approved it but added a small question :P

crate/operator/restore_backup.py
plaharanne: It is to keep in consistent with the other fields of `AwsBackupRepositoryData` that have type `str` and not `Optional[str]`

PR #895, opened 2026-09-01
Set the external-dns hostname annotation under both the alpha and new prefix

WalBeh:
## Summary of changes

external-dns [v0.22.0](https://github.com/kubernetes-sigs/external-dns/releases/tag/v0.22.0) changes the default annotation prefix from `external-dns.alpha.kubernetes.io/` to `external-dns.kubernetes.io/`, with no fallback to the old prefix. The operator now sets the hostname annotation under both prefixes (option 1 in the issue: "two won't hurt") on:

- the data service (`_lb_annotations_to_add`, plus removal via `_lb_annotations_to_remove` when switching to Traefik)
- the grand-central nginx `Ingress`

Both keys live in `constants.py` with a note to drop the alpha one once all external-dns controllers are confirmed on v0.22.0+.

For existing clusters (item 2 in the issue), `devtools/backfill_external_dns_annotation.py` — a uv script (PEP 723 inline metadata, click CLI) — copies the alpha annotation's value to the new key on operator-managed Services and Ingresses only, selected via `app.kubernetes.io/managed-by=crate-operator,app.kubernetes.io/part-of=cratedb`. It dry-runs by default, takes `--apply` to patch and `--context` to pick a kubeconfig context, skips resources that already have the new annotation, and reports per-resource failures without aborting the sweep (non-zero exit if any). Run once per region:

```
uv run devtools/backfill_external_dns_annotation.py --context <ctx>          # dry run
uv run devtools/backfill_external_dns_annotation.py --context <ctx> --apply
```

Still out of scope: pinning `--annotation-prefix` / bumping external-dns on the deployments in kubernetes-gitops (item 3).

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/3080

### DIFF
--- a/devtools/backfill_external_dns_annotation.py
+++ b/devtools/backfill_external_dns_annotation.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env -S uv run --script
+# /// script
+# requires-python = ">=3.12"
+# dependencies = [
+#     "click>=8",
+#     "kubernetes>=31",
+#     "urllib3>=2",
+# ]
+# ///
+"""Backfill the new external-dns hostname annotation on CrateDB resources.
+
+external-dns v0.22.0 changed its default annotation prefix from
+"external-dns.alpha.kubernetes.io/" to "external-dns.kubernetes.io/" with no
+fallback to the old one. The operator sets both prefixes on newly created
+resources; this script adds the new annotation to existing operator-managed
+Services and Ingresses. Run once per Kubernetes cluster. Resources that
+already carry the new annotation are skipped, so re-running is safe.
+"""
+
+import sys
+
+import click
+import urllib3
+from kubernetes import client, config
+from kubernetes.client.rest import ApiException
+
+ALPHA = "external-dns.alpha.kubernetes.io/hostname"
+NEW = "external-dns.kubernetes.io/hostname"
+SELECTOR = "app.kubernetes.io/managed-by=crate-operator,app.kubernetes.io/part-of=cratedb"  # noqa
+
+
+@click.command(help=__doc__)
+@click.option("--apply", is_flag=True, help="Patch the resources (default: dry run).")
+@click.option("--context", help="kubeconfig context to use (default: current).")
+def main(apply: bool, context: str | None) -> None:
+    config.load_kube_config(context=context)
+    # Whether to verify TLS comes from the kubeconfig; don't also warn about it.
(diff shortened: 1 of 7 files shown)


devtools/backfill_external_dns_annotation.sh
```diff
@@ -0,0 +1,62 @@
...
+      continue
+    fi
+    if $APPLY; then
+      kubectl annotate "$kind" -n "$ns" "$name" "$NEW=$value"
+    else
+      echo "would annotate $kind $ns/$name with $NEW=$value"
+    fi
+    patched=$((patched + 1))
+  done < <(
+    kubectl get "$kind" --all-namespaces -o json |
```
goat-ssh: maybe we should skip non-cratedb namespaces, as it now tries to also patch traefik which is managed by fluxcd helmrelease:
```
...
would annotate service f7f39a73-eb11-4e81-8450-f63c214ca9b9/crate-3de31e35-d82c-447a-a36b-b133a63f7818 with external-dns.kubernetes.io/hostname=tan-zam-wesell.aks1.eastus2.azure.cratedb.net.
would annotate service traefik/traefik with external-dns.kubernetes.io/hostname=traefik.aks1.eastus2.azure.cratedb.net.
Would annotate 20, 0 already done. Re-run with --apply.
```

WalBeh: but it would not hurt to set it there also, right? I assume adding the annotation does not re-trigger a helm install.

goat-ssh (approved):

devtools/backfill_external_dns_annotation.py
```diff
@@ -0,0 +1,79 @@
...
+@click.command(help=__doc__)
+@click.option("--apply", is_flag=True, help="Patch the resources (default: dry run).")
+@click.option("--context", help="kubeconfig context to use (default: current).")
+def main(apply: bool, context: str | None) -> None:
+    config.load_kube_config(context=context)
+    core = client.CoreV1Api()
+    networking = client.NetworkingV1Api()
+    patched = skipped = failed = 0
+    for kind, list_fn, patch_fn in [
+        ("service", core.list_service_for_all_namespaces, core.patch_namespaced_service),
```
tomach:
```suggestion
        (
            "service",
            core.list_service_for_all_namespaces,
            core.patch_namespaced_service,
        ),
```

tomach (approved):
LGTM! 💯

PR #902, opened 2026-09-08
Add CrateDB CRD reference page

tomach:
## Summary of changes
Adds a new documentation page, `CrateDB CRD reference` (`docs/source/crd-reference.rst`), and links it into the toctree in `index.rst`.

The operator's `cratedbs.cloud.crate.io` CRD had no human-readable reference. Since we recommend the operator for self-hosted deployments, this fills that gap with a single page describing the important settings and where they go.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/3104


docs/source/crd-reference.rst
```diff
@@ -0,0 +1,733 @@
...
+     hosts (one CrateDB pod per node), so ``replicas: 3`` needs at least three
+     schedulable worker nodes. On a small test cluster, lower ``replicas`` to
+     match the number of nodes you have.
+   * **Resources** — size ``requests``/``limits`` to fit your nodes; the values
+     above are only an example.
+
+The top-level ``spec`` object has the following members. Only ``cluster`` and
+``nodes`` are required.
+
+.. list-table::
```
plaharanne, line 97: would it make sense to make this file a md? Because for example, here, the table is hard to read without the preview mode.

tomach: hm I think I would keep it as .rst. The whole docs tree is .rst, built with Sphinx for read-the-docs, and we don't have the necessary tools installed. so a .md page wouldn't render on the docs site without adding new dependencies and Sphinx extension just for this one file

plaharanne (approved):
💯

PR #905, opened 2026-09-17
Add OpenShift Partner Validation report

tomach:
## Summary of changes

Adds the public artifacts for the Red Hat OpenShift Partner Validation of the operator, plus a few OpenShift docs fixes found while running it.

- `partner-validation-report.md` - test plan + results for the 9 lifecycle scenarios (TC-01–TC-09).
- `RUNBOOK.md` - how to execute the validation, and how to submit via Partner Connect.
- `capture.sh` - captures read-only verification evidence.
- `manifests/` - namespace + CrateDB test-cluster manifests.

All scenarios passed on OpenShift `4.22.11` (EUS), operator `2.64.1`, CrateDB `6.4.4`.

## Checklist

- [x] Link to issue this PR refers to: https://github.com/crate/cloud/issues/2981


openshift-validation/capture.sh
```diff
@@ -0,0 +1,144 @@
...
+  [[ "$1" == "--" ]] && shift
+  {
+    echo "===================================================================="
+    echo "# $title"
+    echo "# \$ $*"
+    echo "# captured: $(date -u '+%Y-%m-%dT%H:%M:%SZ') (UTC)"
+    echo "===================================================================="
+    "$@" 2>&1
+    echo
+    echo "# exit code: $?"
```
WalBeh: Heads-up, this makes the evidence slightly misleading: by the time we reach the last line, `$?` is the exit status of the blank `echo` above it, not of the captured command — so every snapshot records `exit code: 0`, even when the command failed. Capturing it right after the command fixes it:

```suggestion
    "$@" 2>&1
    local rc=$?
    echo
    echo "# exit code: $rc"
```

openshift-validation/capture.sh
```diff
@@ -0,0 +1,144 @@
...
+  } >"$file"
+  echo "  -> $1  ($file)"
```
WalBeh: After the shifts, `$1` is the command (`/usr/bin/oc`), so this prints `-> /usr/bin/oc  (...)` instead of something useful. `$title` is what you want:

```suggestion
  echo "  -> $title  ($file)"
```

openshift-validation/RUNBOOK.md
```diff
@@ -0,0 +1,215 @@
...
+./capture.sh 4.22 cratedb crate-operator
+oc get scc | grep crate-anyuid
+oc get pod <pod> -n cratedb -o jsonpath='{.metadata.annotations.openshift\.io/scc}{"\n"}'
+```
+
+Cluster health (via a Route or `psql`) — record into the evidence folder:
+
+```bash
+oc create route passthrough cratedb-http --service=crate-my-cluster --port=4200 -n cratedb
+# then: SELECT health FROM sys.cluster;   -> expect GREEN
```
WalBeh: `sys.cluster` doesn't have a `health` column (it's the id/name/master/settings table), so this errors with `ColumnUnknownException` if it's pasted as-is. The single-row cluster verdict lives in `sys.cluster_health`:

```suggestion
# then: SELECT health FROM sys.cluster_health;   -> expect GREEN
```

openshift-validation/capture.sh
```diff
@@ -0,0 +1,144 @@
...
+# ---- TC-06: operator upgrade (version snapshot) ---------------------------
+snap "tc-06-operator-image.txt" "TC-06 operator image after upgrade" -- \
+  "$CLI" get deploy crate-operator -n "$OPERATOR_NS" -o jsonpath='{..image}'
+
+echo
+echo "Done. Review the files in $EVIDENCE_DIR and link them from the"
+echo "matching Evidence: slots in partner-validation-report.md."
+echo
+echo "Reminders — capture these manually (not read-only):"
+echo "  * Cluster health:  SELECT health FROM sys.cluster;  -> paste into tc-*.txt"
```
WalBeh: Same query as the RUNBOOK reminder — `sys.cluster` has no `health` column, so this fails if pasted:

```suggestion
echo "  * Cluster health:  SELECT health FROM sys.cluster_health;  -> paste into tc-*.txt"
```

openshift-validation/partner-validation-report.md
```diff
@@ -0,0 +1,573 @@
...
+```console
+# 1. Install CRDs
+$ helm repo add crate-operator https://crate.github.io/crate-operator
+$ helm install crate-operator-crds crate-operator/crate-operator-crds
+
+# 2. Install the operator in openshift mode
+$ helm install crate-operator crate-operator/crate-operator \
+    --set env.CRATEDB_OPERATOR_CLOUD_PROVIDER=openshift \
+    --set env.CRATEDB_OPERATOR_CRATE_CONTROL_IMAGE=crate/crate-control:<tag> \
+    --namespace crate-operator --create-namespace
```
WalBeh, line 176: The report is the artifact reviewers read, so its steps should run as written. This one skips the two flags the runbook (step 2) and the chart README require: without `crate-operator-crds.enabled=false` the second install fails with a CRD ownership error (the operator chart bundles the same CRD chart), and without the debug-volume StorageClass the pods stay `Pending` on an unbound PVC.

```suggestion
$ helm install crate-operator crate-operator/crate-operator \
    --set env.CRATEDB_OPERATOR_CLOUD_PROVIDER=openshift \
    --set env.CRATEDB_OPERATOR_CRATE_CONTROL_IMAGE=crate/crate-control:<tag> \
    --set env.CRATEDB_OPERATOR_DEBUG_VOLUME_STORAGE_CLASS=<your-storageclass> \
    --set crate-operator-crds.enabled=false \
    --namespace crate-operator --create-namespace
```

openshift-validation/RUNBOOK.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+**Prepare for TC-04:** create a test table with a replica so shards exist in
+other zones (needed for the zone-loss resilience sub-test):
+
+```sql
+CREATE TABLE validation.t (id INT, v TEXT)
+  CLUSTERED INTO 6 SHARDS WITH (number_of_replicas = 1);
+INSERT INTO validation.t VALUES (1, 'a'), (2, 'b');
```
WalBeh, line 127: On a fresh cluster only `doc` exists, so `CREATE TABLE validation.t` fails with `SchemaUnknownException[Schema 'validation' unknown]`. Creating the schema first makes the snippet copy-pasteable:

```suggestion
CREATE SCHEMA IF NOT EXISTS validation;
CREATE TABLE validation.t (id INT, v TEXT)
  CLUSTERED INTO 6 SHARDS WITH (number_of_replicas = 1);
INSERT INTO validation.t VALUES (1, 'a'), (2, 'b');
```

openshift-validation/partner-validation-report.md
```diff
@@ -0,0 +1,573 @@
...
+
+**Steps:**
+
+```console
+$ oc new-project cratedb
+$ oc label namespace cratedb \
+    pod-security.kubernetes.io/enforce=privileged \
+    pod-security.kubernetes.io/warn=privileged \
+    pod-security.kubernetes.io/audit=privileged
+$ oc apply -f dev-cluster.yaml   # 3 data nodes, storageClass set to the tested SC
```
WalBeh: `dev-cluster.yaml` doesn't exist in the repo — looks like a leftover from the local run. The committed manifest is `manifests/02-cratedb.yaml`:

```suggestion
$ oc apply -f manifests/02-cratedb.yaml   # 3 data nodes, storageClass set to the tested SC
```

WalBeh:
While you're in this file — the last Limitations bullet is stale:

```rst
- **``CRATE_CONTROL_IMAGE`` is required**: The operator does not validate
  that this variable is set. If it is missing, StatefulSets will be
  created with an empty image reference and pods will fail to pull.
```

`create.py` raises a `kopf.PermanentError` when `CLOUD_PROVIDER=openshift` and the sidecar image is unset ("CRATEDB_OPERATOR_CRATE_CONTROL_IMAGE must be set when ..."), and report §4 describes exactly that fail-fast behaviour. Suggested replacement:

```rst
- **``CRATEDB_OPERATOR_CRATE_CONTROL_IMAGE`` is required**: The operator fails
  fast when ``CLOUD_PROVIDER=openshift`` and the sidecar image is unset, so a
  missing value is caught at deploy time. StatefulSets cannot be created without
  a sidecar image.
```

(The old text also used the bare `CRATE_CONTROL_IMAGE` name, missing the `CRATEDB_OPERATOR_` prefix used everywhere else.) Not introduced by this PR, but you're touching the file and the report names the docs as canonical, so it'd be good to fix here.

---

Three cosmetic RST things in the same file — entirely take-or-leave, docutils doesn't warn on any of them (only mentioning them because `concepts.rst`/`configuration.rst` use exact-length underlines):

- **L186–190**: `crate/crate-control:<version>` is an image reference, not a command, so in a `.. code-block:: console` it looks copy-pasteable. An inline literal reads better: `You can use the official image directly without building it yourself, e.g. ``crate/crate-control:<version>``.`
- **L414**: the only >88-char line in the file — the `oc get pod … jsonpath=…` command wraps cleanly with a `\` like the sibling commands do.
- **L122 / L282 / L308 / L365 / L417**: heading underlines 1–2 chars longer than their titles (harmless, purely house style).

openshift-validation/partner-validation-report.md
```diff
@@ -0,0 +1,573 @@
...
+- all pods `Ready`, CrateDB cluster health **GREEN**, correct node count.
+
+**Verification:**
+
+```console
+$ oc get pods -n cratedb -l app.kubernetes.io/component=cratedb
+$ oc get scc | grep crate-anyuid
+$ oc get pod <pod> -n cratedb -o jsonpath='{.metadata.annotations.openshift\.io/scc}'
+# Cluster health (via Admin UI Route or SQL):
+#   SELECT health FROM sys.health;  -> GREEN
```
WalBeh: Small consistency thing: `sys.health` works, but it returns one row *per table/partition*, so "expect GREEN" is a bit stronger than the query alone shows. `sys.cluster_health` gives the single-row cluster verdict. Same spot in TC-03 (`:259`) and TC-06 (`:397`) — TC-04's variant with the `missing_shards`/`underreplicated_shards` columns is fine as-is.

```suggestion
#   SELECT health FROM sys.cluster_health;  -> GREEN
```

openshift-validation/partner-validation-report.md
```diff
@@ -0,0 +1,573 @@
...
+
+**Result:** ✅ **Evidence:** [metrics](evidence/ocp-4.22/tc-08-observability.txt) · [Admin UI](evidence/ocp-4.22/tc-08-adminui.png)
+
+> **Observed:** each CrateDB pod exposes a JMX/Prometheus metrics endpoint on
+> `:7071/metrics`, returning real metrics (`crate_query_*`, `crate_node` shard
+> info/stats: primaries 2 / replicas 2 / unassigned 0). The Admin UI (via Route/
+> port-forward on 4200) shows the healthy 3-node cluster (version 6.4.4, status
+> GREEN). The SQL-exporter sidecar (`:9399`) provides additional SQL-based
+> metrics and requires an SSL-enabled cluster to connect. It operates normally on an SSL-enabled cluster (the production
+> default). Prometheus observability is validated here via the `:7071` endpoint.
```
WalBeh: The last two sentences here repeat themselves, and they tell the reader the `:9399` exporter needs SSL while the report marks SSL out of scope (§1b) — so the step above reads as if it was confirmed. Something like:

```suggestion
> metrics but requires an SSL-enabled cluster to connect — that is out of scope
> for this validation (see §1b), so it was not exercised here. Prometheus
> observability is validated via the `:7071` endpoint.
```

Would be nice to match step 1 (L440–442) to this, e.g. mark the `:9399` part ➖ N/A — the legend already has that symbol.

openshift-validation/RUNBOOK.md
```diff
@@ -0,0 +1,215 @@
...
+helm upgrade --install crate-operator-crds crate-operator/crate-operator-crds \
+  --namespace crate-operator --create-namespace
+
+# operator, with the bundled CRD subchart disabled
+helm upgrade --install crate-operator crate-operator/crate-operator \
+  --set env.CRATEDB_OPERATOR_CLOUD_PROVIDER=openshift \
+  --set env.CRATEDB_OPERATOR_CRATE_CONTROL_IMAGE=crate/crate-control:2.64.1 \
+  --set env.CRATEDB_OPERATOR_DEBUG_VOLUME_STORAGE_CLASS=standard-csi \
+  --set crate-operator-crds.enabled=false \
+  --namespace crate-operator --create-namespace
```
WalBeh, line 68: Since the report attests operator `2.64.1`, pinning the chart here makes a re-run reproduce that exact version (the tags exist now). Same for the CRDs release above (`--version 2.64.1`); the blockquote below can then drop its `<chart>` placeholder.

```suggestion
helm upgrade --install crate-operator crate-operator/crate-operator \
  --version 2.64.1 \
  --set env.CRATEDB_OPERATOR_CLOUD_PROVIDER=openshift \
  --set env.CRATEDB_OPERATOR_CRATE_CONTROL_IMAGE=crate/crate-control:2.64.1 \
  --set env.CRATEDB_OPERATOR_DEBUG_VOLUME_STORAGE_CLASS=standard-csi \
  --set crate-operator-crds.enabled=false \
  --namespace crate-operator --create-namespace
```

openshift-validation/RUNBOOK.md
```diff
@@ -0,0 +1,215 @@
...
+to refresh the read-only snapshots, and `tee` the action's own output as noted.
+
+---
+
+## Teardown (when done)
+
+```bash
+oc delete cratedb my-cluster -n cratedb
+helm uninstall crate-operator -n crate-operator
+helm uninstall crate-operator-crds
```
WalBeh: Without `-n` this uninstalls from whatever namespace the current context points at; the release lives in `crate-operator` (the line above gets it right):

```suggestion
helm uninstall crate-operator-crds -n crate-operator
```

WalBeh (commented):
LGTM, please have a look at the comments

openshift-validation/RUNBOOK.md
tomach: I think I'll leave this as-is. CrateDB creates the schema implicitly on `CREATE TABLE`, so `validation.t` works on a fresh cluster without a preceding `CREATE SCHEMA`. https://cratedb.com/docs/crate/reference/en/latest/general/ddl/create-table.html#schemas 
adding it here would slightly muddy the TC-09 cleanup.

WalBeh: Right, CrateDB has no CREATE SCHEMA at all, my mistake.

WalBeh (approved):
LGTM. Two non-blocking things: the crate-operator-crds install in RUNBOOK.md isn't pinned to 2.64.1 yet, and the passthrough route can't work without SSL on the cluster (the same command is in openshift.rst on master, so fine as a follow-up).

PR #906, opened 2026-09-22
Harden the STACKIT availability zone lookup

WalBeh:
## Summary of changes

`node.attr.zone` is a command substitution that `bin/crate` evaluates at start-up, so whatever curl prints becomes the zone. `-s` does not suppress the response body and the exit status was unchecked, so an erroring or not yet reachable metadata service made its body the zone. A body with leading whitespace also stops the node from starting.

Adds `-f`, a timeout and retries, and replaces anything that is not a plausible zone with `lookup-failed`. STACKIT only — SKE brings pod networking up late, so a restart races the metadata route. No other provider has reported this.

Only `awk` and `curl` are used, both already in the CrateDB image.

Takes effect on pod restart only; the StatefulSet is `OnDelete`.

## Checklist

- [x] Link to issue this PR refers to: crate/cloud#3121


tomach (approved):
LGTM! for existing clusters the STS would need to be updated manually, right?

